PR #7, opened 2017-07-10
Remove all files, that aren't needed by maven

madoar:
This PR finishes the introduction of maven by removing all other build files.
Now to work with the project in an IDE it's important to import it as a maven project!
In addition you can now run all tests without an IDE in the terminal via:
```bash
mvn clean test
```

To build the application without an IDE you need to run:
```bash
mvn clean package
```
This command will create a `rails-2.0-RC4-jar-with-dependencies.jar` file in the `target` folder.
This file is executable and starts rails after executing.

If we want to add other types of installers in addition, like an `.exe` file for windows or a `.deb` file for debian systems, we can do this by using additional maven plugins.


neutronc:
I would for now not remove the build.xml and build.gradle files. They contain information and build recipes (like the webstart and mac app creation) i am not sure we have yet with maven. Until i have successfully understood maven and used that to build the release files for github i would keep them here.

madoar:
The problem with multiple build systems is, that they always have some differences, which makes it hard to maintain them all at the same time.
One example are the used dependency versions.
The `build.xml` file for example uses some quite old dependency versions.
`batik` for example is used in version `1.7` in the `build.xml` script, when I'm using version `1.9` in maven.

About the webstart application:
Is this currently used? I'm asking, because when I take a look at the release page on sourceforge I see only three files: 

- the jar archive
- an exe file for windows systems
- and a file for mac systems.

If you can give me a complete list of executables we need to build for a release I'll try to add them to the maven setup, so that you can generate everything with a simple `mvn clean package` command.

neutronc:
Hi Marc,

i didnt create the webstart yet, but inside the build.gradle you can see 
the commands to do so.

Actually we have at least 5 different "packages".

One large "standalone Jar" that should be useable on anyplattform.
WIndows:
The executable
A Zipfile containing a smaller JAR and some Batchscript as well as the 
Readmes - Which can also be used on *nix-like Systems

MacOs
a DMG - macbased installer
a ZIP File containing a mac app


My ideal thought would be that each of those will be build and packaged 
seperately with the Readmes with an installer that has been signed by us :)

Regards,
Martin

Am 10.07.17 um 17:47 schrieb madoar:

neutronc:
I'll update the Batik Library :) and the others today and tomorrow.

Am 10.07.17 um 17:47 schrieb madoar:

madoar:
Some questions:
- What is the difference between the smaller jar and the standalone jar, my guess is that the dependencies are missing is this right?
- What is the difference between the DMG installer and the content inside the zip archive?

If you feel more safe with gradle, we can also use gradle for the dependency management and the build process. The problem here is, that I have no prior experience with gradle and can't help with that.
I just want to prevent to add an additional build system nobody later wants to use, therefore we should choose a single build system that we want to use and remove all other systems.

neutronc:
The small jar doesnt have the dependencies, they are sitting in a 
directory "libs" next to the rails jar.
The standalone Jar has everything incoporated.

They are pretty much the same, but the dmg you could codesign with a 
valid developer key wile the zip-archive you cant.

No dont misunderstand me, right now i want to keep the avenues, but 
again as i said i agree with you one tool if it does the same as the 
mess we have now (including the webstart) is the way to go !

Regards,
Martin

Am 10.07.17 um 18:03 schrieb madoar:

madoar:
I've just added two plugins to the maven pom file:
- one plugin for the creation of an `exe` file for windows and
- one plugin for the creation of a `dmg` installer for mac os

I'm currently only using linux. Therefore I'm currently unable to test the created files.
I'm not sure what OS you use @neutronc but can you test one or both of the created files?

madoar:
Ok, I've just tested the `exe` application on a windows 7 system and it works :)

madoar:
Do we really need to provide the mac zip archive and a slim jar archive? 
I'm not sure I understand the benefit of a slim jar archive with a separate dependency folder in a personal context.

madoar:
A little tip from me:
I think it's better to use the merging mode `Squash and merge`.
The reason:
Often you want to view a PR as a single state transition or as a single commit, this can be achieved by using `Squash and merge`.

neutronc:
Again the slim jar was an old thing, if we can get the package together 
with the readmes, the .bat and the .sh im all for dropping it for the 
all-jar

Am 10.07.17 um 20:54 schrieb madoar:

neutronc:
Thanks,
btw are you on slack ?

Am 10.07.17 um 20:59 schrieb madoar:

madoar:
Which slack?

neutronc:
slack.com

try to access rails-18xx.slack.com

Am 10.07.17 um 21:02 schrieb madoar:

neutronc:
https://join.slack.com/rails-18xx/shared_invite/MjA5OTE2NjcwMjc1LTE0OTk2MDEzNTItMTg3MDMyMGFhMQ

Am 10.07.17 um 21:02 schrieb madoar:

PR #19, opened 2017-07-13
Remove eclipse specific files

madoar:
Currently the repository contains quite a lot of eclipse specific configuration files.
These files are sometimes customized to fit to a specific user/developer setup and should therefore not be added to the repository.
In addition these files are auto generated by eclipse when the maven project gets imported.


madoar:
Yes you're right. Somehow I forgot to add the deleted files to `.gitignore`.

Issue #28, opened 2017-07-17
1889 - "null" included in company description strings


o0dano0o:
Hi,

In the map window for 1889 if you click on Info->Companies and then go through each of the public and privates, some of the strings have "null" at the end. It looks like all companies are affected in some way by this.

For example, in private company A, the last line says:
Blocking: K4null

Thanks!

neutronc:
Confirmed as not only tagging 1889. Also 1830 and probably others have that problem.

neutronc:
Fixed by PR #29

PR #29, opened 2017-07-17
Fixes Issue #28 String "null" appearance

neutronc:
Fixes Issue #28 Fixing the apperance of the String null in the infotext on Private andPublic Companies where it didnt belong, stemming from the use of
uninitialised variables.


blentz:
:+1:

Issue #31, opened 2017-07-17
Bug 1835 - Stefan is Prussia director


madoar:
In the appended screenshot you'll discover, that Stefan is asked which train of the Prussia he wants to discard. The problem here is, that Rainer is the Prussia director, not Stefan. 
Our guess is, that Rails got confused, when the private companies were forcefully traded into Prussia shares.
![grafik](https://user-images.githubusercontent.com/18488086/28292086-dee55576-6b4d-11e7-8661-fcd54b0ec568.png)

neutronc:
Actually its not a bug but Stefan is the player currently acting. In the moment in his turn hes triggering the Discard.Train Step in the OR and asked since hes the "currentPlayer" to fullfill that action. Since its a popup you cant break out of it.

PR #34, opened 2017-07-18
Quick adjustment.

neutronc:
#31 Giving the acing player notice to contact the director of the company that needs to discard trains. A different approach might be forthcoming at a later stage.


madoar:
The code looks good.
Can you please add a screenshot that shows how this looks like now?

blentz:
bildschirmfoto 2017-07-18 um 17 22 41

Issue #30, opened 2017-07-17
Bug 1835 - Prussia share buyable


madoar:
When looking at the appended screenshot you'll discover, that Rails says, Rainer can buy a Prussia (PR) share. The problem is, that he doesn't have enough money to do this, 92 < 154. 
Therefore I would expect rails not to highlight the Prussia share.
![grafik](https://user-images.githubusercontent.com/18488086/28291984-6f1b304e-6b4d-11e7-8f7d-9ab856a8ffdb.png)

neutronc:
You have a savegame available?

madoar:
I have two save games:

- The first save game captures the state directly before buying the first `5` train, which forces everyone else to swap their privates to Prussia shares.
- The second save game contains the end state (Containing the results)

I'll append both, the save game with the later time (21:20) is the end result.
[Rails-bugs.zip](https://github.com/Rails-18xx/Rails/files/1155501/Rails-bugs.zip)

PR #36, opened 2017-07-19
Fixing setBuyableShares for multi share Certs

neutronc:
Before this fix a cert with multiple shares was shown as buyable because
th price check did not account for multiple shares per cert. The buying
would have failed neverthe less. With this fix the price per shareunit 
in the price check is taking into account the multipe shares per cert if
they are present. This only works from the IPO currently. Theres a small
chance that the wrong display might appear on shares in the pool (but
very unlikely).
This is a fix for Issue #30

### DIFF
--- a/src/main/java/net/sf/rails/game/financial/StockRound.java
+++ b/src/main/java/net/sf/rails/game/financial/StockRound.java
@@ -263,7 +263,7 @@ public class StockRound extends Round {
 
                 if (!cert.isPresidentShare()) {
                     price = comp.getIPOPrice() / unitsForPrice;
-                    if (price <= playerCash) {
+                    if ((price*shares) <= playerCash) {
                         possibleActions.add(new BuyCertificate(comp, cert.getShare(),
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/rails/game/financial/StockRound.java
```diff
@@ -263,7 +263,7 @@ public void setBuyableCerts() {
 
                 if (!cert.isPresidentShare()) {
                     price = comp.getIPOPrice() / unitsForPrice;
-                    if (price <= playerCash) {
+                    if ((price*shares) <= playerCash) {
```
blentz: Within this block, the "shares" variable is only referenced here, which makes creating the variable unnecessary.

It'd be simpler for this to be "price*cert.getShares()"

neutronc:
I have updated in accordance with your comments the routine in question.

madoar:
@neutronc this PR seems to make some tests fail?

PR #43, opened 2017-07-23
Bring in the corrected Testreports

neutronc:


erik-vos:
Please put a space between the old and new text in the log message.

PR #46, opened 2017-07-23
Add jdeb plugin to pom file to create a debian installer

### LINKED ISSUE #24: Add maven plugin to create .deb installer files
It would be quite nice if we add a maven plugin to our project that creates a `.deb` installer for debian systems.
A maven plugin to create `.deb` installers can be found here: https://github.com/tcurdt/jdeb/blob/master/docs/maven.md

madoar:
Fixes #24


pom.xml
```diff
@@ -280,6 +280,59 @@
...
+                                        <type>perm</type>
+                                        <prefix>/usr/share/applications</prefix>
+                                    </mapper>
+                                </data>
+                                <data>
+                                    <src>${project.build.directory}/${jarName.all}.jar</src>
+                                    <type>file</type>
+                                    <mapper>
+                                        <type>perm</type>
+                                        <prefix>/usr/local/lib/rails</prefix>
```
blentz: This should probably be "/usr/lib/rails-18xx" or similar, so we avoid conflicting with ruby-on-rails.

PR #58, opened 2018-05-17
Created .Net Standard 2.0 port of the non-GUI Rails code

JimWuerch:
Created a .net port of the rails non-gui code.

Load/Save game works, but I'm not sure how much actually works yet.  A demo opening round is included the builds with .Net Core.

The State tests are written with mstest.  They can likely be ported to xUnit as is, but I needed something quick and dirty to start.

There's lots of inconsistencies with Properties and getters/setters now, so a cleanup pass for them is likely in order.  I tried to stick with properties where they made sense, but sometimes the rabbit hole got a bit too deep, or a derived class would do something that might throw.

Likely the namespaces could do with a bit of renaming too, but before that work all happens, I'd like a consensus as to what they should be.


blentz:
Rails is a *Java* project. There is ZERO interest in a port to .NET. You've wasted your time on porting to a language we care nothing about.

There is literally no reason for us to merge this. It doesn't fix any bugs. It doesn't implement anything new. It addresses nothing in our backlog of open issues. It's a complete waste of everyone's time.

Either contribute to the project IN THE LANGUAGE WE ARE USING or move on.

Issue #60, opened 2018-09-26
JavaFX


madoar:
Just a little brainstorming on my side:
I believe it would be useful to change the GUI implementation to JavaFX.
In my opinion JavaFX has multiple advantages:
- JavaFX is a newer framework compared to Swing/AWT, which looks more up to date and is easier to program, especially with tools like the SceneBuilder (see: https://gluonhq.com/products/scene-builder/)
- JavaFX in theory should be relatively easy to port to IOS and Android (see: https://gluonhq.com/products/mobile/javafxports/)

What do you think?

jklap:
@madoar the current `FXStockChartWindow` implementation is acting as a singleton and starting a JavaFX `Application`. Can you please revisit this? My understanding is that you can mix JavaFX and Swing components together-- do we need to create this `Application`? 

If we do then it should be independent of the `FXStockChartWindow` class.

Can you please either:
* remove the use of the `Application` if that is possible (suspect not but thought I'd ask) or
* refactor the code such that the `Application` is independent and separate from the actual `FXStockChartWindow` implementation

I realize that you originally wrote this as a POC to show as an example idea but given this is now becoming the main-line branch think this code needs to be cleaned up to make it production ready

madoar:
It should be possible to integrate a JavaFX component in a Swing `JFrame`. To do this we would need to make use of a `JFXPanel`. For more details see [here](https://docs.oracle.com/javase/8/javafx/interoperability-tutorial/swing-fx-interoperability.htm#CHDIEEJE) and [here](https://docs.oracle.com/javase/8/javafx/interoperability-tutorial/simpleswingbrowserjava.htm#CBBHEAII).

I'm currently not sure when I will have the time to work on this. It may take a while.

erik-vos:
See also issue #297. JavaFX is not for me.

neutronc:
![grafik](https://user-images.githubusercontent.com/6752457/100721831-23113600-33c0-11eb-8029-fbfd9678e5b3.png)

neutronc:
Is that what you wanted Erik ?

erik-vos:
Yes, that is great.

PR #61, opened 2018-09-28
Change the Swing stock chart implementation to a JavaFX implementation

madoar:
This PR represents the first step for #60.
It introduces a new stock chart window written in JavaFX, which scales the size of the window.

After the rewrite, the functionality to save and restore the location and size of the window has been temporarily removed.

Feel free to give the PR a test to see if something needs to be improved.

Some screenshots:
grafik
grafik
grafik


madoar:
This reminds me:
JavaFX may not be available by default on all systems:
- OracleJDK 10-: on the OracleJDKs up to version 10 JavaFX is installed by default, which is the reason why the first test succeeds 
- OpenJDK: 10-: on OpenJDK for linux you need to install OpenJFX separately, I don't know if this is required on windows or mac.
   On linux, this becomes a problem with Java 9 and 10, because the developers changed the way OpenJFX needs to be installed (it must be somehow built together with OpenJDK).
   I'm not really sure if there are any prebuilt packages of Java 9 and Java 10 out there for linux, on which OpenJFX works.
- Java 11: for Java 11, which has been released at the beginning of this week, the situation looks different. Starting with Java 11, JavaFX can be added to an application through maven.

madoar:
@blentz travis doesn't seem to provide the `openjfx` package, which contains the JavaFX dependencies.
An alternative is using a docker container, but I'm not sure if this is to much at this point.
What do you think?

madoar:
Ok, this is interesting:
- I've now added all java versions, which are compatible with rails.
- in addition, it seems like the plugin for the generation of an osx installer doesn't work with Java 10 and 11, so I've removed it for now. I'm not sure how we want to handle this problem long term.

PR #65, opened 2018-10-20
Rails 2 maintenance

neutronc:
Fixing an erroneous behaviour in the 1856 CGR Formation Round.


neutronc:
Ok i have to fix the test  reports first i think. i am going to reject that PR.

Issue #76, opened 2019-10-26
1835 - wrong player gets PD if the last action was a sell action


peter-de-boer:
According to the rules, the player left to the player who did the last buy action gets the PD.

In the implementation, if the last action was a sell action though, the player left to the seller gets the PD. 

See the attached example. Peter did the last action (sell), and Pep got the PD. But according to the rules, Hakko (left to Dan who was the last buyer) should get PD.
[1835_20191026_1505_Dan.zip](https://github.com/Rails-18xx/Rails/files/3774956/1835_20191026_1505_Dan.zip)

erik-vos:
No. The original AH rulebook states in rule 4.0: "When everyone has 
stopped trading the player on the left of the last player to buy or sell 
is given the Priority Card (...)".
You are referring to rule 7.0, which deals with the first stock round 
only, in which it is not allowed to sell any stock.

Erik

Op 26-10-2019 om 17:11 schreef peter-de-boer:

peter-de-boer:
No, I am referring to the original German rules (Hans im Glück, 1990). 
Rule 2.3 says:

"Der Spieler links von
demjenigen Spieler, der zuletzt eine Kaufaktion getätigt hat. bekommt die
Holzlokomotive (ist Startspieler der nächsten AR)."

erik-vos:
Sorry, please disregard my previous message, in which I was referring to 
1830. Indeed 1835 seems to differ, a deviation from 1830 that has 
escaped my attention so far. That also is the cause that Rails has not 
implemented it. The 18xx rules difference list indeed reports this 
difference for 1835 and a few other games.

My bad.
Erik


Op 26-10-2019 om 17:11 schreef peter-de-boer:

neutronc:
closed with latest commit to Repository.

PR #88, opened 2020-03-16
Fix for Issue #76

neutronc:
Refactored setPriority in StockRound with an argument indicating a kind of Action. Action currently cant derived directly..
More refactoring necessary perhaps at a later stage. But in the Moment the setPriority Routine gets overwritten in 1835 and 1825 for priority changes only after Buy actions not after sell or other actions.
Check needed for StartCompany...


src/main/java/net/sf/rails/game/specific/_1825/StockRound_1825.java
```diff
@@ -81,5 +81,14 @@ protected void gameSpecificChecks (PortfolioModel boughtFrom, PublicCompany comp
...
     }
 
+	@Override
+	protected void setPriority(String string) {
+		if (string.equals("BuyCert")) {
+		super.setPriority(string);
+		}
+		else {
+			return;
+		}
```
madoar: This part is not required

src/main/java/net/sf/rails/game/specific/_1825/StockRound_1825.java
```diff
@@ -81,5 +81,14 @@ protected void gameSpecificChecks (PortfolioModel boughtFrom, PublicCompany comp
 
     }
 
+	@Override
+	protected void setPriority(String string) {
+		if (string.equals("BuyCert")) {
+		super.setPriority(string);
+		}
```
madoar: Formatting

src/main/java/net/sf/rails/game/specific/_1835/StockRound_1835.java
```diff
@@ -271,5 +271,15 @@ protected boolean checkIfSplitSaleOfPresidentAllowed() {
...
+
+	@Override
+	protected void setPriority(String string) {
+		if (string.equals("BuyCert")) {
+		super.setPriority(string);
+		}
+		else {
+			return;
+		}
+	}
```
madoar: See above

src/main/java/net/sf/rails/game/specific/_1837/StockRound_1837.java
```diff
@@ -208,7 +208,7 @@ protected boolean mergeCompanies(PublicCompany minor, PublicCompany major) {
                 setSellObligationLifted(major);
             }
 
-            setPriority();
+            setPriority("MergeCompany");
```
madoar: Maybe replace with an enum?

src/main/java/net/sf/rails/game/specific/_1825/StockRound_1825.java
neutronc: technically you are correct...

src/main/java/net/sf/rails/game/specific/_1837/StockRound_1837.java
neutronc: Quicker to be handled than the string mechanics ?

madoar: What do you mean by `quicker`? In my opinion it is more secure. When you use a `String` you have the possible issue of a misspelling, enums don't allow for such errors.

In theory it should also be possible to not do these checks at all by using a class comparison which is really ugly...

neutronc:
I'll have a look in the Stockround SetPriority handling tomorrow...
And probably redo that. If anyone else in the meantime has spare time you are more than welcome to do like madoar suggests..

PR #92, opened 2020-03-19
Add GitHub Actions

madoar:
- execute the tests on JDK 8
- execute the tests on JDK 11

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,35 @@
+name: Java CI with Maven
+
+on:
+  push:
+    branches: [ rails_2_develop ]
+  pull_request:
+    branches: [ rails_2_develop ]
+
+jobs:
+  jdk8:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+        java-package: jdk+fx
+    - name: Tests - OpenJDK 8 
+      run: mvn -B package --file pom.xml
+  jdk11:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


madoar:
@neutronc this PR makes use of Github Actions instead of Travis to run the tests and build the `jar` archive. The benefit here is that we can run the tests on all three most popular operating systems and with both JDK 8 and 11. What do you think about this (especially compared to Travis)?

neutronc:
Hi what ever helps us with automatic testing is good :)

Btw i think i am a bit nearer to the error with the game that fails in tests.
On Windows it works as a charm on linux it bails in the JavaFX section of trying to build a Configuration ? So what ever is the error is treated in JavaFX differently on Linux than on Windows.

madoar:
Why do you think the issue is in the JavaFX code?

neutronc:
Cause it bailed on my Ubuntu system in the ConfigWindow.java ? 
Check it out by loading the 1851 test game ?

madoar:
But the failing test is for an 1835 test case?

neutronc:
grafik

On my windows system they all run with the current base without any issues..

madoar:
Maybe it is really something about the game files that are different on windows compared to other systems

madoar:
Ok maybe I can help you somehow. I debugged the test case a bit. 

Let's start at the end. The test fails at:

https://github.com/Rails-18xx/Rails/blob/841adc14a974d2afea52d252e417a205851c103b/src/main/java/net/sf/rails/util/GameLoader.java#L271

This occurs because previously `GameManager#processOnReload` returns `false` in:

https://github.com/Rails-18xx/Rails/blob/841adc14a974d2afea52d252e417a205851c103b/src/main/java/net/sf/rails/game/GameManager.java#L880-L884

This again is the case because `possibleActions#validate` returns `false` in:

https://github.com/Rails-18xx/Rails/blob/841adc14a974d2afea52d252e417a205851c103b/src/main/java/rails/game/action/PossibleActions.java#L86-L102

The internal state of the method/object when it returns `false in` line 101 is as followed:

grafik

If you need additional information feel free to ask :)

PR #94, opened 2020-03-19
Remove Travis CI

madoar:
Now that we have added Github Actions we can remove Travis CI because it is redundant.


madoar:
@neutronc I think it is a good idea to change the project settings so that a PR can only be merged when all tests are green. (I can't do that myself because I don't have the necessary permissions)

madoar:
If you want we can also run the tests on changes in the maintenance branch

neutronc:
> 
> 
> @neutronc I think it is a good idea to change the project settings so that a PR can only be merged when all tests are green. (I can't do that myself because I don't have the necessary permissions)

I have raised your access level

madoar:
Thanks. I couldn't find the setting, but if it is fine for you I will remove all travis specific options from the settings now that we have removed the travis configuration file?

PR #98, opened 2020-03-24
refresh tile counts in RemainingTileWindow

jklap:
Changes to the RemainingTileWindow so that it updates tiles left without having to close and re-open the window during real-time play


jklap:
Looks like it's due to cherry-picking a commit for the PR. Going to kill this one until the earlier ones are pushed

PR #100, opened 2020-03-24
refresh tile counts in RemainingTileWindow

jklap:
Remaining tile counts shown in window don't refresh automatically during real-time play. This fixes that issue

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/RemainingTilesWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/RemainingTilesWindow.java
@@ -6,6 +6,8 @@ import java.awt.FlowLayout;
 import java.awt.Image;
 import java.awt.event.*;
 import java.awt.image.BufferedImage;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import javax.swing.*;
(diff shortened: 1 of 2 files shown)


jklap:
Yes, tested myself with several manual test cases and also provided a test jar to two others (that had originally complained and brought the initial issue to my attention).

My only concern with this fix is the large number of Observers that end up created -- would have been nice if you could register a single Observer multiple times but that isn't allowed ;)

jklap:
Note too, this change only effects the remaining tile display-- so even if the numbers ended up off it would still be no worse/different then it was previously :)

src/main/java/net/sf/rails/ui/swing/RemainingTilesWindow.java
```diff
@@ -97,28 +101,49 @@ private void init(GameUIManager gameUIManager) {
...
+                    refreshCounts();
+                }
+
+                @Override
+                public Observable getObservable() {
+                    return null;
+                }
+            };
+            tile.getTilesLaid().addObserver(watcher);
+            observerMap.put(tile, watcher);
```
madoar: Off-topic:
It would be nice if we could replace these `Observer` objects with lambda methods.
What are your thoughts?

jklap: I'm all for that! 

I've also converted some code to streams in the SonarQube changes but didn't want to convert much due to not knowing the history

madoar: I think it would be good to have a proof of concept PR showcasing an example what can be changed. If no negative feedback is returned we could continue. 

I'm also not that knowledgeable about the code so it is hard to say what can and should be changed.

jklap: I'm very very far from knowledgable about this code base-- I've only started looking at it a week or so ago!

I'll create a PR this afternoon with a couple of "sample" changes for review and feedback

PR #102, opened 2020-03-24
initial pass at Discord

jklap:
Initial pass at support Discord notifications at the end of a user's turn by supporting the configuration of a Discord Webhook and a user mapping from player name to Discord User Id for @ notifications.

This adds a configuration pane to the existing configuration window.

As these details (both webhook and users) are game specific these configuration options are *not* saved with a user's other configuration choices to disk but instead written to the game file itself. When a game file is loaded the options are written into the Config object for use the same way any other Config option is retrieved.

I fully expect feedback on this PR before it will be accepted, please review at your convenience.


jklap:
Note too, this also adds Config.set() but this only has a transient setting, these changes are not saved with a user's profile (as they are retained) in ConfigurationManager and not pushed down to ConfigProfile. This means if/when a profile is changed is does not change these settings. I had considered flushing the transient cache if the profile was changed but made a decision that it's more likely the settings should carry over. Just a thought for your consideration.

madoar:
I would be interested in knowing how you intend to use this PR? Are you putting your save-game in a central place (for example a shared folder) from which every player can load the game and save it to after making his turn?

jklap:
That is exactly what we are doing (with over 2,000 saved game files/turns thus far). We've experimented with using a Google Drive folder mounted on your desktop using Google Backup-- not so great for real-time play as the sync process can have some delays (1-2 minutes). We are currently using a webdav share mounted to the desktop which has native support in both Windows and Mac OS. I've also heard Dropbox syncing is pretty fast. This PR, along with that shared folder along with use of autoload/polling allows for the play of a game in pretty near to real-time in addition to allowing people to step away for moments and get a notification when their turn rolls around to them.

madoar:
Interesting idea.

Maybe we can build on top of this and build a similar solution in directly into Rails?
For example we could add an alternative `Save` button that send the game state to a `git`-repository or some rest-endpoint. In turn we could provide add a related `Load` button.

Depending on the "backend" we use we could even add an automatic reload mechanism to rails that automatically reloads the game when a change has occurred.

jklap:
I'd started to refactor the GameSave and GameLoad logic into a GamePersistence layer with pluggable persistence classes for just that purpose (was actually thinking of starting out with classes for writing to a remote MySql database and querying an updates as that would be very fast to implement although obviously not very scalable for everyone else to use) but ran into lots of cross dependencies due to all the UI logic in GameUIManager. Paused the work for the time being but have not given up. Obviously once that initial refactor has been completed we could easily add pluggable classes for talking to a REST API or a git repo or an sftp drop.

neutronc:
Very nice, that might also solve the problem i currently ran into.
Creating randomized start and bidding packages is currently not possible  due to the current reconfiguration from start for each load.

PR #103, opened 2020-03-24
initial pass at Slack

jklap:
Slack notification support similar to Discord notification support.

Added Slack webhook and user name mapping configurations to existing Notification config pane with the Discord configurations


jklap:
Something like a Config.infoText kind of thing or documentation elsewhere?

neutronc:
I honestly think we need both :)

jklap:
I'll add the Config.infoText.

Where should the "elsewhere" be?

neutronc:
Hm thinking out loud, it really belong into our Wiki imho.
Otherwise we would have to have a stack of docs being created and delivered.

But maybe we can discuss that in our slack channel for more input from other people ?

PR #105, opened 2020-03-25
addl improvements to polling

jklap:
Changes to polling to:
* Check for the saved game file before passing it to GameUIManager which can occur over a shared folder where the last rails file is synced before the actual game file
* Decreased the polling interval while we wait for the game file to show up
* Added a game saved file length check (backward compatible) so that we don't pick the file up and pass to the GameUIManager before the full file is written out. This still needs changes to support in the writing of the last rails file. This scenario shouldn't happen often

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
+++ b/src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
@@ -13,91 +13,102 @@ import rails.game.action.GameAction;
 
 
 public class AutoLoadPoller extends Thread {
-    
-    private GameUIManager guiMgr;
+
+    private final GameUIManager guiMgr;
     private String saveDirectory;
     private String savePrefix;
     private String ownPostfix;
     private int pollingInterval;
     private int pollingStatus;
-    
+
     private boolean pollingActive = false;
-    
-    private String lastSavedFilenameFilepath;
+
+    private final String lastSavedFilenameFilepath;
     private String lastSavedFilename = "";
-    
+
     public static final int OFF = 0;
     public static final int ON = 1;
     public static final int SUSPENDED = 2;
 
-    protected static Logger log =
-        LoggerFactory.getLogger(AutoLoadPoller.class);
+    protected static Logger log = LoggerFactory.getLogger(AutoLoadPoller.class);
 
     public AutoLoadPoller (GameUIManager guiMgr, String saveDirectory, String savePrefix, String ownPostfix,
             int status, int pollingInterval) {
- 
         this.guiMgr = guiMgr;
         this.saveDirectory = saveDirectory;
         this.savePrefix = savePrefix;
         this.ownPostfix = ownPostfix;
         this.pollingStatus = status;
         this.pollingInterval = pollingInterval;
-        
+
         lastSavedFilenameFilepath = saveDirectory + "/" + savePrefix + ".last_rails";
-        
-        log.debug("Poller own postfix: "+ownPostfix);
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
```diff
@@ -13,91 +13,102 @@
...
         log.info ("AutoLoadPoller started");
 
+        int currentPollInterval = 1;
         int secs, sleepTime;
-        String currentFilename;
 
         for (;;) {
+            secs = Calendar.getInstance().get(Calendar.SECOND);
+            try {
+                sleepTime = 1000 * (currentPollInterval - secs%currentPollInterval);
```
madoar: You can move the variable declaration inside this line

src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
```diff
@@ -13,91 +13,102 @@
...
     @Override
     public void run () {
-
         log.info ("AutoLoadPoller started");
 
+        int currentPollInterval = 1;
         int secs, sleepTime;
-        String currentFilename;
 
         for (;;) {
```
madoar: Off-topic: should we use `for (;;)` or `while (true)`? Or can we maybe use some kind of scheduler?

src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
```diff
@@ -13,91 +13,102 @@
...
+                    if ( ! currFile.exists() ) {
+                        log.debug("Saved file {} missing, waiting", currFile);
+                        currentPollInterval = 1;
+                        continue;
+                    }
+
                     if (!lastSavedFilename.equals(currentFilename)) {
+                        if ( fileSize != null ) {
+                            long fileSizeNum = Long.parseLong(fileSize);
+                            if ( currFile.getTotalSpace() != fileSizeNum ) {
```
madoar: Maybe shorter: `if (fileSize != null && Long.parseLong(fileSize) != currFile.getTotalSpace())`

This would allow us to remove the inner loop

src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
jklap: I didn't touch any of the existing code unless it was related to the actual changes needed-- I've run into lots of possible changes of this nature but haven't touched them unless I would have touched the actual line in question. Note that the SonarQube PR that I'll be pushing in does start to make these types of changes but man, there are a LOT

src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
jklap: Personally I never ever use `for (;;)`, I always use `while (true)` although realistically we should be checking a flag and shutting the poller down on exit so this case shouldn't really be either of these!

src/main/java/net/sf/rails/ui/swing/AutoLoadPoller.java
jklap: Yes and makes the code harder to read and for the brain to parse given the 3 conditions in the `if ()`. In most cases I prefer the code as written as it is not only easier to read and understand but also because collapsing it down to a single line doesn't buy us anything-- the only time I tend to allow for harder to read code is if there is a strong benefit such as some type of optimization or performance -- but even those are questionable changes with the compilers of today. You would also then have a second hit on getting the file sizes for the logging statement a couple of lines down that you don't have.

PR #106, opened 2020-03-25
code cleanup

jklap:
First set of code cleanups from SonarQube are about to PR’ed. These first ones should have min logic changes or even real code changes-- its a lot of log parameterization, some minor merging of lines that had been split, import cleanups


jklap:
Self rejecting due to a merge error

PR #109, opened 2020-03-26
code samples

jklap:
Sample code "improvements" bringing the code base to 2020 :)

As request for feedback by the team by @madoar

### DIFF
--- a/src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
+++ b/src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
@@ -112,9 +112,9 @@ class RevenueCalculatorMulti extends RevenueCalculator {
         }
 
         // finished all tries
-        for (int i=0; i < startVertexes.length; i++) {
+        for ( int startVertex : startVertexes ) {
             // remove all of them from the visited vertex list
-            trainVisited[trainId][startVertexes[i]] = false;
+            trainVisited[trainId][startVertex] = false;
         }
 
         // allow that the train does not run at all
--- a/src/main/java/net/sf/rails/game/Bonus.java
+++ b/src/main/java/net/sf/rails/game/Bonus.java
@@ -2,13 +2,13 @@ package net.sf.rails.game;
 
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import net.sf.rails.algorithms.NetworkVertex;
 import net.sf.rails.algorithms.RevenueAdapter;
 import net.sf.rails.algorithms.RevenueBonus;
 import net.sf.rails.algorithms.RevenueStaticModifier;
-
-
+import net.sf.rails.game.state.Observable;
 
 /**
  * An object of class Bonus represent extra income for the owning company,
--- a/src/main/java/net/sf/rails/game/GameManager.java
+++ b/src/main/java/net/sf/rails/game/GameManager.java
@@ -253,12 +253,16 @@ public class GameManager extends RailsManager implements Configurable, Owner {
                             skipFirstStockRound);
 
                 for (String ruleTagName : srTag.getChildren().keySet()) {
-                    if (ruleTagName.equals("NoSaleInFirstSR")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_IN_FIRST_SR, true);
-                    } else if (ruleTagName.equals("NoSaleIfNotOperated")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_IF_NOT_OPERATED, true);
-                    } else if (ruleTagName.equals("NoSaleOfJustBoughtShare")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_OF_JUST_BOUGHT_CERT, true);
+                    switch ( ruleTagName ) {
+                        case "NoSaleInFirstSR":
+                            setGameParameter(GameDef.Parm.NO_SALE_IN_FIRST_SR, true);
+                            break;
+                        case "NoSaleIfNotOperated":
+                            setGameParameter(GameDef.Parm.NO_SALE_IF_NOT_OPERATED, true);
+                            break;
+                        case "NoSaleOfJustBoughtShare":
+                            setGameParameter(GameDef.Parm.NO_SALE_OF_JUST_BOUGHT_CERT, true);
+                            break;
                     }
 
                 }
@@ -1200,10 +1204,8 @@ public class GameManager extends RailsManager implements Configurable, Owner {
         List<String> b = new ArrayList<String>();
 
         /* Sort players by total worth */
-        List<Player> rankedPlayers = new ArrayList<Player>();
-        for (Player player : getRoot().getPlayerManager().getPlayers()) {
-            rankedPlayers.add(player);
-        }
+        List<Player> rankedPlayers = new ArrayList<>();
+        rankedPlayers.addAll(getRoot().getPlayerManager().getPlayers());
         Collections.sort(rankedPlayers);
 
         /* Report winner */
--- a/src/main/java/net/sf/rails/game/PlayerManager.java
+++ b/src/main/java/net/sf/rails/game/PlayerManager.java
@@ -281,11 +281,7 @@ public class PlayerManager extends RailsManager implements Configurable {
 
        final boolean ascending_f = ascending;
 
-       Comparator<Player> cashComparator =  new Comparator<Player>() {
-           public int compare (Player p1, Player p2) {
-               return ascending_f ? p1.getCash() - p2.getCash() : p2.getCash() - p1.getCash();
-           }
-       };
+       Comparator<Player> cashComparator = (p1, p2) -> ascending_f ? p1.getCash() - p2.getCash() : p2.getCash() - p1.getCash();
 
        playerModel.reorder(cashComparator);
 
--- a/src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
+++ b/src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
@@ -5,6 +5,7 @@ import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
+import java.util.stream.IntStream;
 
 import javax.swing.*;
 
@@ -125,12 +126,7 @@ public class AutoSaveLoadDialog extends JDialog implements ActionListener {
 
     public void actionPerformed(ActionEvent arg0) {
         if (arg0.getSource().equals(okButton)) {
-            for (int i = 0; i < NUM_OPTIONS; i++) {
-                if (choiceButtons[i].isSelected()) {
-                    status = i;
-                    break;
-                }
-            }
+            status = IntStream.range(0, NUM_OPTIONS).filter(i -> choiceButtons[i].isSelected()).findFirst().orElse(status);
             interval = (Integer) intervalSpinner.getValue();
         } else if (arg0.getSource().equals(cancelButton)) {
             //status = -1;
--- a/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
+++ b/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Observable;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.stream.IntStream;
 
 import javax.swing.*;
 import javax.swing.plaf.FontUIResource;
@@ -775,12 +776,7 @@ public class GameUIManager implements DialogOwner {
                 boolean[] exchanged = dialog.getSelectedOptions();
                 String[] options = dialog.getOptions();
 
-                int numberSelected = 0;
-                for (int index=0; index < options.length; index++) {
-                    if (exchanged[index]) {
-                        numberSelected++;
-                    }
-                }
+                int numberSelected = (int) IntStream.range(0, options.length).filter(index -> exchanged[index]).count();
 
                 int minNumber = action.getMinNumberToExchange();
                 int maxNumber = action.getMaxNumberToExchange();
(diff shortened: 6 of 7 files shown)


src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
```diff
@@ -112,9 +112,9 @@ protected void runTrain(final int trainId) {
         }
 
         // finished all tries
-        for (int i=0; i < startVertexes.length; i++) {
```
jklap: Use of enhanced `for()` loops vs us of index based iteration

src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -253,12 +253,16 @@ public void configureFromXML(Tag tag) throws ConfigurationException {
                             skipFirstStockRound);
 
                 for (String ruleTagName : srTag.getChildren().keySet()) {
-                    if (ruleTagName.equals("NoSaleInFirstSR")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_IN_FIRST_SR, true);
-                    } else if (ruleTagName.equals("NoSaleIfNotOperated")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_IF_NOT_OPERATED, true);
-                    } else if (ruleTagName.equals("NoSaleOfJustBoughtShare")) {
-                        setGameParameter(GameDef.Parm.NO_SALE_OF_JUST_BOUGHT_CERT, true);
+                    switch ( ruleTagName ) {
```
jklap: Use of the new `switch()` format that allows String `cases`

src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -1200,10 +1204,8 @@ public boolean getGameOverReportedUI(){
         List<String> b = new ArrayList<String>();
 
         /* Sort players by total worth */
-        List<Player> rankedPlayers = new ArrayList<Player>();
-        for (Player player : getRoot().getPlayerManager().getPlayers()) {
-            rankedPlayers.add(player);
-        }
+        List<Player> rankedPlayers = new ArrayList<>();
+        rankedPlayers.addAll(getRoot().getPlayerManager().getPlayers());
```
jklap: Use of new array operations such as `addAll()` vs iteration

src/main/java/net/sf/rails/game/PlayerManager.java
```diff
@@ -281,11 +281,7 @@ public Player reorderPlayersByCash (boolean ascending) {
 
        final boolean ascending_f = ascending;
 
-       Comparator<Player> cashComparator =  new Comparator<Player>() {
-           public int compare (Player p1, Player p2) {
-               return ascending_f ? p1.getCash() - p2.getCash() : p2.getCash() - p1.getCash();
-           }
-       };
+       Comparator<Player> cashComparator = (p1, p2) -> ascending_f ? p1.getCash() - p2.getCash() : p2.getCash() - p1.getCash();
```
jklap: Use of a lambda to replace anonymous classes

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
```diff
@@ -125,12 +126,7 @@ private GridBagConstraints constraints(int gridx, int gridy, int leftinset,
 
     public void actionPerformed(ActionEvent arg0) {
         if (arg0.getSource().equals(okButton)) {
-            for (int i = 0; i < NUM_OPTIONS; i++) {
-                if (choiceButtons[i].isSelected()) {
-                    status = i;
-                    break;
-                }
-            }
+            status = IntStream.range(0, NUM_OPTIONS).filter(i -> choiceButtons[i].isSelected()).findFirst().orElse(status);
```
jklap: Use of streams with logical operations (filtering) to pull out specific items (in this case first)

src/main/java/net/sf/rails/ui/swing/GameUIManager.java
```diff
@@ -775,12 +776,7 @@ public void dialogActionPerformed (boolean ready) {
                 boolean[] exchanged = dialog.getSelectedOptions();
                 String[] options = dialog.getOptions();
 
-                int numberSelected = 0;
-                for (int index=0; index < options.length; index++) {
-                    if (exchanged[index]) {
-                        numberSelected++;
-                    }
-                }
+                int numberSelected = (int) IntStream.range(0, options.length).filter(index -> exchanged[index]).count();
```
jklap: Steams to filter & count

jklap:
All JDK 1.8 supported.

src/main/java/net/sf/rails/game/Bonus.java
```diff
@@ -2,13 +2,13 @@
...
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import net.sf.rails.algorithms.NetworkVertex;
 import net.sf.rails.algorithms.RevenueAdapter;
 import net.sf.rails.algorithms.RevenueBonus;
 import net.sf.rails.algorithms.RevenueStaticModifier;
-
-
+import net.sf.rails.game.state.Observable;
```
madoar: Why are these imports added when nothing else changed?

src/main/java/net/sf/rails/game/GameManager.java
madoar: Essentially you're writing here mapping logic between a `String` and an `GameDef.Parm` enum.
I think it would be better to move this mapping logic to a new 

```java
public Parm parse(String ruleTagName) {
   ...
}
```

method in the `Parm` enum.
Afterwards we could use it like: `setGameParameter(GameDef.Parm.parse(ruleTagName), true);`

src/main/java/net/sf/rails/game/GameManager.java
madoar: Maybe even better:

```java
List<Player> rankedPlayers = getRoot().getPlayerManager().getPlayers().stream().sorted().collect(Collectors.toList());
```

src/main/java/net/sf/rails/game/PlayerManager.java
madoar: Here I would recommend to use:

```java
Comparator<Player> cashComparator = Comparator.comparing(Player::getCash);
```

If you need a comparator for the other direction you can then use:

```java
cashComparator.reversed()
```

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
madoar: This is fine in my opinion, but if we need to add more steps to the stream the expression becomes hard to read, therefore I would suggest to add some newlines to the expression. For example like this:

```java
status = IntStream.range(0, NUM_OPTIONS)
   .filter(i -> choiceButtons[i].isSelected())
   .findFirst().orElse(status);
```

src/main/java/net/sf/rails/game/GameManager.java
neutronc: This chained call is really hard to read and probably to understand..
Of course i am no formally educated programmer...

src/main/java/net/sf/rails/game/PlayerManager.java
neutronc: interesting, when did that :: operater got implemented ?

src/main/java/net/sf/rails/game/GameManager.java
jklap: These are only examples of newer JDK 8 syntax changes, not actual code rewrites which this type of enum parsing would be :)

src/main/java/net/sf/rails/game/GameManager.java
jklap: I also find chained calls hard to read and I am a formally trained developer! That is why I have formatting rules set up to automatically reformat something like the above into something like this:
```
List<Player> rankedPlayers = getRoot().
    getPlayerManager().
    getPlayers().
    stream().
    sorted().
    collect(Collectors.toList());
```
Streams (or chains actually) are initially a bit more difficult to understand but once you learn to mental "parse" them they actually can be much easier to understand than the equiv written out verbosely.

In this case though the use of `addAll()` which is cleaner is likely more appropriate than a `stream()` as the original code had no sorting, ie we are just looking at these examples as direct equivalences simply for example purposes.

src/main/java/net/sf/rails/game/PlayerManager.java
jklap: @neutronc it was added as in 8 to help with `stream()` and lambdas and is a reference to a method, static, object, etc. Check section 3 here: https://www.baeldung.com/java-8-new-features

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
jklap: Yep completely agree, see my above about use of automatic formatting. I didn't format any of the example here as I use automatic formatting and that would have changed the whole darn file and no one would have been able to find the example in all of the format changes!

src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
```diff
@@ -112,9 +112,9 @@ protected void runTrain(final int trainId) {
...
+        for ( int startVertex : startVertexes ) {
```
jklap: Use of enhanced for() loops vs us of index based iteration

src/main/java/net/sf/rails/game/Bonus.java
jklap: Because by accident I also pulled this file out of the set of changes while pulling the other examples. Nothing to see here ;)

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
jklap: Another note-- this is a good example why I think it's important that we have formatting rules "embedded" into the project-- I've got a set of IntelliJ files that I'll be PRing that include code style formatting rules so that everyone that is using IntelliJ will get the same set of rules. Would be nice if there was a way we could convert those rules to the Eclipse code formatter and add those files to the repo as then anyone using Eclipse OR IntelliJ would apply the same. I know I did the opposite a long long time ago when I migrated from Eclipse to IntelliJ but that was a by-hand project.. ie I took a file that had been formatted with my Eclipse rules, opened it in IntelliJ and started tweaking the rules one-by-one until executing code format in IntelliJ made no changes to the file.

src/main/java/net/sf/rails/game/GameManager.java
madoar: > In this case though the use of addAll() which is cleaner is likely more appropriate than a stream() as the original code had no sorting, ie we are just looking at these examples as direct equivalences simply for example purposes.

But the code has sorting, see:

https://github.com/jklap/Rails/blob/4937f172666931c233b6518cee8183cf8608e18a/src/main/java/net/sf/rails/game/GameManager.java#L1207-L1209

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
madoar: The question is where and how should we apply these rules? I think the best approach is to use a linter that checks for violations of the codestyle rules.

src/main/java/net/sf/rails/game/GameManager.java
jklap: This is why I hate GitHubs code review tool-- too easy to miss the code below the comments!

Yes, either of these solutions work-- and not really sure one is any better than the other. I tend to use the non-stream version here and use `addAll()` and then `Collections.sort()` because I find it easier for others to understand but that is just a personal preference and that in this case both are completely viable solutions

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
jklap: You are referring to code formatting specifically? That is why I build these rules into IntelliJ as rejecting a PR post submission because of style is a waste of everyones time-- build the formatting into the developers process instead of as an afterthought that then get rejected and becomes a painful fix-it-later problem.

BUT even with that manual focus mentioned, yes, adding in a tool to check for not only style violations but also coding/logic violations is a key component to keeping the code base clean and maintainable. Many of the projects I work on make use of SonarQube for this exact scenario and that will automatically will reject a PR it fails a certain threshold. Keep in mind though that many tools, Sonar, CheckStyle, FindBugs, PMD, etc won't have rules for all of code conventions that you can configure in IDEs like Eclipse and IntelliJ so some stuff will always slip through hence why it's important to get a tool in place like the IDE with the rules setup so a developer just gets in the habit of hitting the format button when they save.

I ran Rails through my instance of SonarQube (not SonarCloud as it doesn't support some of the automation and plugins I use and does not have the deeper rule set I prefer) which is where much of these changes came from. BUT I also started to set up this repo in SonarCloud but realized this morning I don't have project permissions here to trigger a build-- another weird SonarCloud thing. In any case, it would be worth while pursuing this automated scanning setup and feeding it into a PR even if we don't use it to reject/accept...

neutronc:
I installed the SonarCloud and i am going to raise your access level.

src/main/java/net/sf/rails/ui/swing/AutoSaveLoadDialog.java
madoar: > You are referring to code formatting specifically? That is why I build these rules into IntelliJ as rejecting a PR post submission because of style is a waste of everyones time

I agree it would be best if everyone would use their IDE to format his/her code before commiting. The problem is that there are a lot of people that simply don't do that. Even worse there are also a not too small number of people that don't use a classic IDE but instead some other kind of editor like VS Code. I know you can also configure code formatting for such editors but every supported editor/IDE requires its own set of configuration, which makes it unfeasible. 

Another reason for rejecting a PR based on style violations is that you can often only get hold of a developer until their PR has been merged. Afterwards they often move on, either to another issue or to another project. So if you don't want to refactor your code after every three or four merged PRs then you should push your developers to provide clean, well formatted and documented code.

jklap: Take a look at the Sonar changes PR

PR #111, opened 2020-03-27
Creating a persistent seed for randomization purposes

neutronc:
First implementation of a mechanismn to automatically create a seed and make it accessible to different code sections.


madoar:
The idea is to first fetch a random seed during game creation. This can be done for example with:

```java
final long seed = new Random().nextLong();
```

This seed should be added to the save file. 

I don't know how the game state is loaded. If it is loaded by "replaying" the previous games from the performed actions you can then make use of deterministic "randomness" for example with:

```java
// init a new Random object during load
final Random rnd = new Random(seed);

// shuffle a list
Collections.shuffle(collection, rnd);

// get a random index
final int index = rnd.nextInt();
```

The idea behind the `seed` is that two `Random` objects will behave the same if you use the same `seed` parameter, i.e. `new Random(42).nextInt() == new Random(42).nextInt()`

If the game state is loaded as a single state then we will need to add some more complex logic.

PR #112, opened 2020-03-27
Sonar cleanup

jklap:
First pass at fixes identified by SonarQube

### DIFF
--- a/src/main/java/net/sf/rails/algorithms/RevenueBonusTemplate.java
+++ b/src/main/java/net/sf/rails/algorithms/RevenueBonusTemplate.java
@@ -17,7 +17,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
-/** 
+/**
  * defines a template for a revenue bonus at creation time of rails objects
  * will be converted to a true RevenueBonus object during each revenue calculation
  * @author freystef
@@ -32,16 +32,16 @@ public final class RevenueBonusTemplate implements Configurable {
 
     // bonus name
     private String name;
-    
+
     // template condition attributes
     private final List<Integer> identVertices;
     private final List<String> identTrainTypes;
     private final List<String> identPhases;
 
     public RevenueBonusTemplate() {
-        identVertices = new ArrayList<Integer>();
-        identTrainTypes = new ArrayList<String>();
-        identPhases = new ArrayList<String>();
+        identVertices = new ArrayList<>();
+        identTrainTypes = new ArrayList<>();
+        identPhases = new ArrayList<>();
     }
 
     public void configureFromXML(Tag tag) throws ConfigurationException {
@@ -53,7 +53,7 @@ public final class RevenueBonusTemplate implements Configurable {
         if (vertexTags != null) {
             for (Tag vertexTag:vertexTags) {
                 Integer id = vertexTag.getAttributeAsInteger("id");
-                if (id != null) { 
+                if (id != null) {
                     identVertices.add(id);
                 }
             }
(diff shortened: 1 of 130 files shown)


madoar:
This is an automated cleanup right?

jklap:
Kinda-- some of the changes are "automated" by IntelliJ but actioned by me (see IntelliJ "intentions" if curious what I mean by this)-- ie specifically picking what should be fixed vs IntelliJ just fixing everything it finds. So vetted. The changes chosen should be low risk, ones that are higher risk (such as class hier duplicated variable names) I left alone if I couldn't trace the full change implications.

src/main/java/net/sf/rails/algorithms/RevenueBonusTemplate.java
```diff
@@ -64,24 +64,24 @@ public void configureFromXML(Tag tag) throws ConfigurationException {
...
-                if (type != null) { 
+                if (type != null) {
                     identPhases.add(type);
                 }
             }
         }
-        log.info("Configured " + this);
-        
+        log.debug("Configured {}", this);
+
```
madoar: Why does it change the log level?

src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
```diff
@@ -154,38 +154,38 @@ final protected void runBottom(final int trainId) {
...
+        log.debug("RC: finished bottom of {}", trainId);
 
 
     }
 
     private final void nextVertex(final int trainId, final int vertexId) {
 
         // 1. encounterVertex adds value and returns true if value vertex
-        Terminated trainTerminated = Terminated.NotYet; 
+        Terminated trainTerminated = Terminated.NOT_YET;
```
madoar: Which Styleguide are you using for this rename?

jklap: Constants being uppercase is a pretty standard convention-- but in this case to answer your question specifically, in this case it's part of the standard checkstyle rules and out-of-the-box SonarCloud/SonarAnalyzer rule sets. 

But even more so, the code base already had a mix of upper case constants and lower or mixed case so in reality the change here isn't so much about a particular style guide but to bring a standard into our own code and given that the bulk of the constant declarations were already upper case it make sense to continue with that approach.

src/main/java/net/sf/rails/algorithms/RevenueBonusTemplate.java
jklap: Because this level of logging detail should match what the rest (or most) of the application is using for log levels. Most of the code is using debug for the lower level, developer useful information (as well it should) and info/warn/error for information that would be helpful for a non-developer. The detail being logged here seemed to fall into the category of information more relevant to a developer reviewing log data than to an end user and hence the level change. There are some pretty consistent standards on use of log levels and while the bulk of the code base does seem to follow them there was a chunk of code that seemed out of place to what the rest of the code was doing-- the bulk of that was this level of `info` to `debug`. So again, while there are standards, in this case it was more about bringing about consistency (or following existing consistency) within the code base.

src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
madoar: I'm fine with this change, but in my opinion `_` as a divider between words looks strange in Java. I know this pattern mostly from C/C++

src/main/java/net/sf/rails/algorithms/RevenueBonusTemplate.java
madoar: :+1:

src/main/java/net/sf/rails/algorithms/RevenueCalculatorMulti.java
jklap: Yep, understood. Sometimes I really miss writing C & C++ :)

PR #114, opened 2020-03-29
Use jpackage to build a native executable/installer

madoar:
This is still WIP


jklap:
Hi @madoar -- why would we build the apps/dmg/deb/etc as part of this GitHub flow vs part of the gradle build? This isn't reusable outside of GitHub-- while I think we should use the flow to publish the artifacts I think they should be built within gradle

madoar:
@jklap I think you misunderstand me. 

This is a proof of concept at the moment (therefore the draft PR). In the I want to have one or more dedicated `tasks` in our gradle build that build the native installers by executing `jpackage`. Such tasks have the additional benefit that they can access build relevant information like the version number and use them during the `jpackage` execution.

The main reason why I didn't implement it as such in this PR is because our current gradle build is broken and your gradle build in #116 is not completed/merged yet. So I have no gradle build which I could modify/extend.

jklap:
@madoar -- great, sounds like we are on the same long-term page!

I spent some time this morning working on getting jpackage into the gradle build including incorporating your other comments. Running into an issue given the recentness with JDK 14, gradle 6.3 and the plugin to support jpackage.

madoar:
> I spent some time this morning working on getting jpackage into the gradle build including incorporating your other comments. Running into an issue given the recentness with JDK 14, gradle 6.3 and the plugin to support jpackage.

Do we need a plugin to use `jpackage` in gradle? I think it would be enough to do something like:

```gradle
task createDebInstaller(type:Exec) {
   commandLine 'jpackage',  '--verbose',
     '--name', 'rails-18xx',
     '--icon', './src/main/resources/build/rails.png',
     '--app-version', '2.1.0',
     '--input', 'target',
     '--main-jar', 'rails-2.2-alpha-all.jar',
     '--linux-package-name', 'rails-18xx',
     '--linux-app-category', 'games',
     '--linux-shortcut'

   //store the output instead of printing to the console:
   standardOutput = new ByteArrayOutputStream()

   //extension method stopTomcat.output() can be used to obtain the output:
   ext.output = {
      return standardOutput.toString()
   }
}
```

jklap:
@madoar can you look into setting up file associations, ie:

> File associations
> 
> You can define one or more file-type associations for your application via the --file-associations option, which can be used more than once. The argument to this option is a properties file with values for one or more of the following keys:
> 
> extension specifies the extension of files to be associated with the application,
> mime-type specifies the MIME type of files to be associated with the application,
> icon specifies an icon, within the application image, for use with this association, and
> description specifies a short description of the association.

madoar:
What file association do you want to have?

jklap:
".rails" for certain. Should probably have ".last_rails" too.

madoar:
I can give it a try. Do you by chance know what happens when a file association triggers an application. Does it pass the filename to the application? Or does something else happen?

madoar:
@jklap I added a `file-association` for `.rails` files. At least on Linux Mint it is working as intended.

To continue integrating `jpackage` I think we first need to finalize the Gradle build file. Can you share your current state of the Gradle build file?

PR #116, opened 2020-03-29
bring gradle build up to date

jklap:
See issue 115


build.gradle
```diff
@@ -1,114 +1,147 @@
...
-  id 'com.github.johnrengelman.shadow' version '5.0.0'
-}
-
-// Addition of macAppBundle
-// https://github.com/crotwell/gradle-macappbundle
 plugins {
+  id 'java'
+  id 'maven-publish'
+  // https://github.com/johnrengelman/shadow
+  id 'com.github.johnrengelman.shadow' version '5.2.0'
```
madoar: This is for the creation of fat-jars right?
I'm currently not sure whether we should take this direction.
`jpackage` seems to be able to work with a collection of multiple jars, i.e. a `lib` folder where all the dependencies are located

build.gradle
```diff
@@ -1,114 +1,147 @@
...
+  // plugin to check for dependency updates via task dependencyUpdates
+  // https://github.com/ben-manes/gradle-versions-plugin
+  id 'com.github.ben-manes.versions' version '0.28.0'
```
madoar: I would recommend to use services like [dependabot](https://dependabot.com/) instead

build.gradle
```diff
@@ -1,114 +1,147 @@
...
+  // https://github.com/n0mer/gradle-git-properties
+  id 'com.gorylenko.gradle-git-properties' version '2.2.2'
+  // https://github.com/TheBoegl/gradle-launch4j
+  id 'edu.sc.seis.launch4j' version '2.4.6'
```
madoar: If we decide to go with `jpackage` I would recommend to remove this plugin to prevent redundancy

build.gradle
```diff
@@ -1,114 +1,147 @@
...
+  // https://github.com/crotwell/gradle-macappbundle
   id "edu.sc.seis.macAppBundle" version "2.3.0"
```
madoar: If we decide to go with `jpackage` I would recommend to remove this plugin to prevent redundancy

build.gradle
```diff
@@ -1,114 +1,147 @@
...
+  // https://github.com/nebula-plugins/gradle-ospackage-plugin
+  //id "nebula.rpm" version "8.2.0"
+  //id "nebula.deb" version "8.2.0"
+  id "nebula.ospackage" version "8.2.0"
```
madoar: If we decide to go with `jpackage` I would recommend to remove this plugin to prevent redundancy

build.gradle
```diff
@@ -1,114 +1,147 @@
...
-apply plugin: "edu.sc.seis.launch4j"
-
-apply plugin: 'com.admc.javaPropFile'
+publishing {
+    publications {
+        maven(MavenPublication) {
+            from(components.java)
+        }
+    }
+}
```
madoar: What does this do?

build.gradle
```diff
@@ -1,114 +1,147 @@
...
+    testImplementation 'junit:junit:4.13'
+    testImplementation 'org.easytesting:fest-assert-core:2.0M10'
+    testImplementation 'org.easytesting:fest-util:1.2.5'
+    testImplementation 'org.mockito:mockito-core:3.3.3'
+}
 
+javafx {
+    version = "14"
+    modules = [ 'javafx.controls' ]
+}
```
madoar: We should also check how to include all native jars (i.e. for linux, mac and windows) inside our `jar` bundles. Instead of limiting it to the one for the current os. I'm not sure if the `org.openjfx.javafxplugin` plugin does this automatically.

build.gradle
```diff
@@ -1,114 +1,147 @@
...
 
-dependencies {
-  compile fileTree(dir: 'libs/main', include: "**/*.jar")
-  testCompile fileTree(dir: 'libs/test', include: "**/*.jar")
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
```
madoar: Maybe better `JavaVersion.VERSION_1_8`

jklap:
This was actually just for discussion purposes, I don't know that it's ready for actual merging given the concern I raised in the related issue regarding JDK 8 vs 11+

build.gradle
jklap: Worth looking into but it seems dependabot's focus is on automating dependency updates vs this reporting on it? It would seem that we have a place for both of these?

build.gradle
jklap: Used to publish the artifacts to maven central. Don't know how we publish today so this likely would be changed

build.gradle
jklap: It would need to be a part of the bundling process. See section 4 here:
* https://github.com/openjfx/javafx-gradle-plugin

build.gradle
jklap: Sure, but convention-wise you'll find that most gradle build script declare it as it is-- easier to read and switching to the constant doesn't really buy us anything. But again, doesn't really matter to me!

build.gradle
madoar: When a new dependency version is released dependabot opens a new PR with the necessary update. Afterwards the maintainers of the project have the choice to either merge the PR or reject it. I think it is quite a good approach.

build.gradle
madoar: As far as I know we don't publish the artifacts to a repository. The most we do currently is to link them on the rails website: https://rails.sourceforge.io/

Long-term I would like to use the `Releases` page on GitHub to host our releases. Maybe even add a GutHub Page as a website.

build.gradle
madoar: My reason for the suggestion is because it is safer. When specifying a number you as a developer need to know what is valid otherwise you can get strange runtime errors. When you use an enum or a static field instead you have compile-time checking that detects wrong java versions (e.g. `1.11` vs `11`)

madoar:
> I don't know that it's ready for actual merging given the concern I raised in the related issue regarding JDK 8 vs 11+

Can you elaborate a bit more about the reason for the issue?
With `maven` you don't have any issues in that regard because the Java 11 libs should be ignored when you run the application on Java 8.

build.gradle
neutronc: The sourceforge release is based upon legacy expectations management. I released the last two builds also on Github. But since most ppl in the past looked for Sourceforge we need to publish the packages on both sites.

On a different Note. Please guys also use the Slack Channel as not all guys interested in Rails development are current following the issues and pull requests here :)

jklap:
Maven isn't doing anything special. The dependencies declared in the pom were all built by JDK 8 or less which is why JDK 8 and JDK 11 work. The difference is the inclusion of the native JavaFX libraries-- those dependencies are built against something higher than JDK 8 and hence won't compile with JDK 8. We could look at switching to the same JavaFX dependency the pom is using but will need to look at what is needed to build in native support

build.gradle
jklap: Was thinking about this more-- given we should really be building artifacts from CI, maybe we don't really need to worry about this? Maybe set up CI to build the artifact for a respective platform FROM that respective platform? What platform options can we build on? Guessing Mac is tough but what about the others?

build.gradle
jklap: Hmmm.. had replied to this but guess I didn't hit submit.

I'm not very familiar with `jpackage` so bear with me and please correct any misunderstandings. But based on your comment it sounds like it's able to bundle up all dependencies but stores them in a directory tree vs generating a single output file? If that is correct then I'd suggest we still need to be able to generate a fat jar. We can't assume that someone, especially someone technical in nature (which I suspect happens to be a large chunk of the 18xx player population!) is going to want to use an app or rpm. And the overhead and supporting that additional generation is minor in nature.

build.gradle
madoar: @neutronc I don't seem to have a link to the Slack channel. Can you please post one? In addition I think it would be good to add the link to the README as well. Maybe behind a nice Button image

build.gradle
madoar: > Maybe set up CI to build the artifact for a respective platform FROM that respective platform?

That is exactly the idea I was going for with #114 ;)

> What platform options can we build on? Guessing Mac is tough but what about the others?

We can build on Windows, Mac and Ubuntu. `jpackage` supports the formats: 

```
    Linux: deb and rpm
    macOS: pkg and dmg
    Windows: msi and exe
```

(as taken from https://openjdk.java.net/jeps/343)

build.gradle
madoar: > We can't assume that someone, especially someone technical in nature (which I suspect happens to be a large chunk of the 18xx player population!) is going to want to use an app or rpm.

He does not need to, he could also load the external dependencies using a `-classpath`/`-cp` parameter when calling `java -jar rails.jar`. Ideally we would wrap this in a shell-/bash-script.

Still this is only an idea. We could also go with a far-jar, I'm not sure how both approaches compare against each other.

PR #120, opened 2020-04-06
add logger to each class. static/final cleanup

jklap:
This PR cleans up the loggers making them private static final and adding several missing loggers (that were instead using loggers up the class hier.

This PR also includes changes making class variables final where obvious.

This PR also removes several class static variables in advance of a larger PR to remove singleton classes/instances.

### DIFF
--- a/src/main/java/net/sf/rails/algorithms/NetworkBuilder.java
+++ b/src/main/java/net/sf/rails/algorithms/NetworkBuilder.java
@@ -39,35 +39,35 @@ package net.sf.rails.algorithms;
 //
 //public class NetworkBuilder implements Iterable<NetworkVertex> {
 //
-//    protected static Logger log =
+//    private static final Logger log =
 //        LoggerFactory.getLogger(NetworkBuilder.class);
-//    
+//
 //    private final SimpleGraph<NetworkVertex, NetworkEdge> mapGraph;
-//    
+//
 //    private final Map<String, NetworkVertex> mapVertexes;
-//    
+//
 //    private NetworkIterator iterator;
-//    
+//
 //    private NetworkBuilder() {
 //        mapGraph = new SimpleGraph<NetworkVertex, NetworkEdge>(NetworkEdge.class);
 //        mapVertexes = new HashMap<String, NetworkVertex> ();
 //    }
-//    
+//
 //    public static NetworkBuilder create(GameManager gameManager) {
 //        NetworkBuilder graphBuilder = new NetworkBuilder();
 //        graphBuilder.generateGraph(gameManager.getMapManager(), gameManager.getRevenueManager());
 //        return graphBuilder;
 //    }
-//    
+//
 //
 //    public void generateGraph(MapManager mapManager, RevenueManager revenueManager) {
-//        
+//
 //        for (MapHex hex:mapManager.getHexesAsList()) {
 //            // get Tile
 //            Tile tile = hex.getCurrentTile();
-//            
+//
 //            // then get stations
-//            Collection<Station> stations = tile.getStations(); 
+//            Collection<Station> stations = tile.getStations();
 //            // and add those to the mapGraph
 //            for (Station station: stations) {
 //                NetworkVertex stationVertex = new NetworkVertex(hex, station);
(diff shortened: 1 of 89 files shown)


jklap:
Some background on the use of private static final for loggers:
* https://owasp.org/www-community/vulnerabilities/Poor_Logging_Practice
* https://rules.sonarsource.com/java/tag/convention/RSPEC-1312

The other changes related to instance variables being static:
* https://cocoacasts.com/are-singletons-bad
* https://www.davidtanzer.net/david%27s%20blog/2016/03/14/6-reasons-why-you-should-avoid-singletons.html

Singletons (or use of static fields for the context of this conversation) aren't inherently bad and certainly have valid uses BUT historically have often been used as a shortcut to the oft difficult problem of making sure a particular class has access to all the other classes it needs. Historically this was solved by making use of Singletons to easily fetch an instance vs passing a slew of objects around. More recenty this has been solved in large apps today by use of IoC or dependency injection (with Spring being one of the most well known DI frameworks).

In our case, we have heavy reliance on the RailsRoot object but also had additional singleton or static variables. This PR does not change the use of RailsRoot as a singleton (that is the coming PR I mentioned) but does reduce or eliminate the use of other singletons or static variables.

This PR also includes changes where a variable can be made final as this offers both the possibility of compiler base optimizations but also makes clear in the code variable usage.

src/main/java/net/sf/rails/algorithms/NetworkBuilder.java
```diff
@@ -39,53 +39,53 @@
 //
 //public class NetworkBuilder implements Iterable<NetworkVertex> {
```
madoar, line 40: Should we perhaps remove this class @neutronc?

src/main/java/net/sf/rails/game/MapOrientation.java
```diff
@@ -126,15 +126,15 @@ public String getUIClassName() {
...
     private static final int[] rowDeltaNS =
             new int[] { +2, +1, -1, -2, -1, +1 };
-    private static final int[] colDeltaNS = 
+    private static final int[] colDeltaNS =
             new int[] {  0, -1, -1,  0, +1, +1 };
-    private static final int[] rowDeltaEW = 
+    private static final int[] rowDeltaEW =
             new int[] { +1,  0, -1, -1,  0, +1 };
     private static final int[] colDeltaEW =
             new int[] { -1, -2, -1, +1, +2, +1 };
```
madoar, line 140: Perhaps change to single line definitions? I.e.:

```java
    private static final int[] rowDeltaNS = new int[] { +2, +1, -1, -2, -1, +1 };
    private static final int[] colDeltaNS = new int[] {  0, -1, -1,  0, +1, +1 };
    private static final int[] rowDeltaEW = new int[] { +1,  0, -1, -1,  0, +1 };
    private static final int[] colDeltaEW = new int[] { -1, -2, -1, +1, +2, +1 };
```

src/main/java/net/sf/rails/game/specific/_1856/CGRFormationRound.java
```diff
@@ -575,7 +572,7 @@ private void formCGR () {
             // First collect old names per city
             Map<String, String> oldTokens = new HashMap<String, String>();
             String cityName;
-            for (BaseToken token : nonHomeTokens) {
+            for (BaseToken token : nonHomeTokens ) {
```
madoar, line 575: Why the empty space here?

src/main/java/net/sf/rails/game/MapOrientation.java
jklap: Done-- will commit as part of the next chunk of code cleanup

src/main/java/net/sf/rails/game/specific/_1856/CGRFormationRound.java
jklap: Typo? Either way, I've started to run the classes through a code formatter so we have same style everywhere but not in a place to commit and push that-- not with other, more important changes happening

src/main/java/net/sf/rails/algorithms/NetworkBuilder.java
neutronc: Thats not really my code but @freystef  i really am not deep in that code there for the network calculation, route and revenue. I can debug it and am after years now familiar with the modifiers... But otherwise if you know something thats easier to read and understand and produces reliable results for network, route and revenue feel free :)

jklap: FYI, this class/file is completely commented out. Maybe that is what @madoar is referring to?

madoar: Exactly :)

PR #121, opened 2020-04-06
bring gradle build up to date

jklap:
Misc tweaks to the gradle build:
* String cleanup (gstring to String)
* Set source and target to JDK 13
* Remove publishing task
* Update gradle wrapper (6.2.2 to 6.3 -- needed for JDK 14)


build.gradle
```diff
@@ -101,12 +93,12 @@ dependencies {
...
 javafx {
-    version = "14"
+    version = '14'
     modules = [ 'javafx.controls' ]
 }
 
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
+sourceCompatibility = JavaVersion.VERSION_13
+targetCompatibility = JavaVersion.VERSION_13
```
madoar, line 101: We need `JavaVersion.VERSION_11`. 

Java 11 is the latest supported long-term version of Java.
Java 13 is not supported for ~3 weeks now.
Java 14 will be supported for the next 6 months.

We should support at least one LTS version of Java

madoar: I don't think that most of us are able to develop and build Rails locally with Java 13+

jklap: Just going by the previous comment made regarding N+1 support, which would be JDK 14 and 13. Doesn't matter to me.

@madoar why are you not able to dev and build locally with JDK 13? Is it a non-US thing? Guessing some type of export restriction or something?

jklap: I'll downgrade that support to 11 for the next merge as this one has already been merged.

madoar: Nothing of the sort. It is about ease of accessibility. On windows I guess it does not make a huge difference. But I'm using Linux Mint. My OS came with Java 11 preinstalled, if you want something newer you need to add another repository that contains the newer version. In addition Java versions that are not LTS version should be changed when their support ends. Otherwise you get no longer updates which is something you don't want.

Summarized it is a question of how easy is it for a user to get the correct Java version to build and run the application himself/herself? The normal user I believe will download a LTS version, i.e. 8 or 11. Only a minority will download Java 14.

jklap: I suspect most users will actually just download the first JDK they see-- not the LTS. Usually you see LTS usage only for businesses that can't/don't want to continually update. If you type "JDK" in Google and go to the JDK page it lists 14 as the "latest release" which is what I'd expect a user to grab vs 11 which is further down the page. But that is for users that would install it manually such as Windows and Mac-- your point to those running off of Linux based releases whether Debian, Mint, Ubuntu (maybe, they tend to stay current) or RedHat/CentOS will certainly be behind. Getting a recent JDK installed on RedHat is a pain and with scl a bigger pain to use (ironically).

In other words, I think if we were developing enterprise level software than LTS would be a requirement. Here, not quite so sure.

But with all of the above said, I'll go with whatever (which FYI is JDK 8, 10, 11, 13 & 14 on my Mac!)

PR #125, opened 2020-04-06
added developer menu to dump logs

jklap:
This PR added a developer menu to trigger a dump of the last 1,000 log files into a file where the saved game file exists.

We have been using this while playing auto-load/save games with a shared folder to dump the logs when one of the players runs into an issue. The log file is then named and written next to the currently/previously loaded game file.

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
+++ b/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
@@ -4,6 +4,8 @@ import java.awt.*;
 import java.awt.datatransfer.Clipboard;
 import java.awt.datatransfer.StringSelection;
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/rails/ui/swing/GameUIManager.java
```diff
@@ -1267,7 +1275,37 @@ public void run() {
...
+        CyclicBufferAppender<?> buffer = (CyclicBufferAppender<?>) lc.getLogger(Logger.ROOT_LOGGER_NAME).getAppender("buffer");
+        if ( buffer == null ) {
+            log.warn("Unable to find cyclic buffer appender");
+            return;
+        }
+
+        PatternLayout layout = new PatternLayout();
+        layout.setContext(lc);
+        layout.setPattern("%d{MM-dd-yyyy:HH:mm:ss.SSS} [%thread] %-5level %logger{10}->%method\\(\\):%line - %msg%n");
+        layout.start();
```
madoar, line 1290: What is the reason that we can't use our "default" approach `private static final Logger log = LoggerFactory.getLogger(GameUIManager.class);` here?

jklap: Because we're not actually creating a logger-- we are fetching a reference to the logger factory, getting a specific appender configured in logback.xml named "buffer" (which is a circular buffer), and then creating a pattern layout to format the log data (in theory you could configure the pattern in a dummy appender in logback.xml and use that but not sure of the value in that).

Finally we iterator over each of the log events currently contained in the buffer appender and write them out to a file.

`Logger log = LoggerFactory...` creates a new log appender for use in generating log events. Our code here is for reading previously created log events and flushing them to a new file on the fly.

PR #131, opened 2020-04-10
Code Refactoring

madoar:
This PR performs some refactoring

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,12 @@
             <version>3.9</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.12</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -183,13 +183,6 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.9</version>
         </dependency>
-
-        <dependency>
-            <groupId>org.projectlombok</groupId>
-            <artifactId>lombok</artifactId>
-            <version>1.18.12</version>
-            <scope>provided</scope>
-        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/net/sf/rails/algorithms/RevenueManager.java
+++ b/src/main/java/net/sf/rails/algorithms/RevenueManager.java
@@ -25,16 +25,15 @@ import org.slf4j.LoggerFactory;
  */
 public final class RevenueManager extends RailsManager implements Configurable {
 
-    private static final Logger log =
-        LoggerFactory.getLogger(RevenueManager.class);
+    private static final Logger log = LoggerFactory.getLogger(RevenueManager.class);
 
     // Modifiers that are configurable
     private final HashSet<Configurable> configurableModifiers = new HashSet<Configurable>();
 
     // Variables to store modifiers (permanent)
-    private final ArrayListState<NetworkGraphModifier> graphModifiers = ArrayListState.create(this, "graphModifiers");
-    private final ArrayListState<RevenueStaticModifier> staticModifiers = ArrayListState.create(this, "staticModifiers");
-    private final ArrayListState<RevenueDynamicModifier> dynamicModifiers = ArrayListState.create(this, "dynamicModifiers");
+    private final ArrayListState<NetworkGraphModifier> graphModifiers = new ArrayListState<>(this, "graphModifiers");
+    private final ArrayListState<RevenueStaticModifier> staticModifiers = new ArrayListState<>(this, "staticModifiers");
+    private final ArrayListState<RevenueDynamicModifier> dynamicModifiers = new ArrayListState<>(this, "dynamicModifiers");
     private RevenueCalculatorModifier calculatorModifier;
 
     // Variables that store the active modifier (per RevenueAdapter)
@@ -56,7 +55,7 @@ public final class RevenueManager extends RailsManager implements Configurable {
         List<Tag> modifierTags = tag.getChildren("Modifier");
 
         if (modifierTags != null) {
-            for (Tag modifierTag:modifierTags) {
+            for (Tag modifierTag : modifierTags) {
                 // get classname
                 String className = modifierTag.getAttributeAsString("class");
                 if (className == null) {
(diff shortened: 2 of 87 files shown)


jklap:
I am personally not a fan of Project Lombok-- while I understand the approach and supposed benefits I feel as though it actually makes things more difficult for a developer to understand and to follow code flows along with making it more likely things are going to get missed especially when doing things like code refactoring, troubleshooting and using tools such as IDEs and code analysis. I find it's better to be explicit and personally don't find the overhead of 1) creating that explicit code difficult (especially given that most IDEs today can create getters and setters with ease) and 2) most developers are used to working with and seeing things like the getters and setters.

Now with that said, I do understand that frameworks such as Spring Boot (which gets lots of attention these days) has introduced some annotations to the world (such as @Slf4s) and do consider those more acceptable but, for instance, the majority of your Java developers (and hence, ie, the people that we'd like to get involved in helping out the project) are not going to understand either of these:
```
@Builder(setterPrefix = "with")
@AllArgsConstructor(access = AccessLevel.PRIVATE)
```

In a nutshell, if this was a corporate project or a personal one I think making a decision to use Project Lombok is perfectly fine but given that we are an open source project and want to encourage new developer involvement (such as the two that recently offered to help) then I think we need to avoid Lombok and stick with more mainstream approaches.

madoar:
For me there are multiple advantages:
- our source code files are shorter. Sometimes even significantly shorter, for example our model classes which contain a lot of final fields.
- because our files are shorter it is much easier to find the significant parts of the code. For example a simple getter method like:

   ```java
   public int getValue() {
      return value;
   }
   ```
   
   does not contain much important information but having ten such methods in a class makes it hard to discover a not so simple getter method like:

   ```java
   public int getComplexValue() {
      if (value != -1) {
         return value;
      } else {
         return defaultValue;
      }
   }
   ```

   Therefore it becomes much easier to detect the significant parts of the code

I agree that we don't necessarily need to use all features of lombok (like the constructor or equals and hashcode generation), but things like getter and setter methods which are auto-generated in 99\% of the cases can be replaced by annotations in my opinion. It does not make a big difference whether you have a `getX` method somewhere below in your class or a `@Getter` before your `x` field.

jklap:
I understand (and have used Lombok on other projects), but working with more junior level developers in those projects I found it to be more problematic than the benefits it provided. In one project we actually ended up making the decision that it caused more understandability (is that a real word?) problems except for the primary app leads and backed it out. I can tell from your comments like the one below that you have worked with it quite a bit and so have internalized it but most developers don't have that, now-for-you, deep understanding and I think it just raises the learning curve for non-senior/professional developers to just jump into this project. Given that vs the (minor in my mind) benefits it provides, I'd still suggest we not pursue it at this time.

> It does not make a big difference whether you have a getX method somewhere below in your class or a @Getter before your x field in detecting that field x has an accessor method.

madoar:
> I can tell from your comments like the one below that you have worked with it quite a bit and so have internalized it but most developers don't have that, now-for-you, deep understanding and I think it just raises the learning curve for non-senior/professional developers to just jump into this project.

Actually I haven't worked with lombok that much. Instead I have worked with other languages (`scala`, `javascript`, `typescript`, `C#`) that come with a subset of the features provided by lombok out of the box. Therefore I'm quite comfortable with them. 

I understand that quite a lot of the features provided by lombok can make the code very hard to understand. I have removed the worst of them (all usages of `@XArgsConstructor` and `@Builder`) from this PR. I think this makes it easier to read.

A problem I see with adding getter and setter method and logger instances by hand or via your IDE is that this approach is hard to maintain and bad on the "reading experience" of the developers/users. As described above having a lot of getter and setter methods in a class makes it easy to miss the significant parts of the class. Adding and maintaining the `log` fields in your class is also error-prone because it can quite easy happen that you use the logger belonging to another class.

> In a nutshell, if this was a corporate project or a personal one I think making a decision to use Project Lombok is perfectly fine but given that we are an open source project and want to encourage new developer involvement (such as the two that recently offered to help) then I think we need to avoid Lombok and stick with more mainstream approaches.

I don't think this is a reason for not using Project Lombok in our codebase. In my opinion the codebase is very complex as it is, especially parts like the calculation of the revenue of the companies during the operation rounds. I think that the right features of lombok can make the code even easier to understand for new developers.

> But then someone really would need to write an introduction to the code or a style guide for this project ?

I think we need such a guide independent of this PR. It would be really useful if we define a set of features that we want to use in the codebase and a formatting that needs to be applied. For me it is quite hard to add code to a project for which I don't understand how the code is formatted and what features of the programming language are allows/recommended by the project maintainers.

neutronc:
> I think we need such a guide independent of this PR. It would be really useful if we define a set of features that we want to use in the codebase and a formatting that needs to be applied. For me it is quite hard to add code to a project for which I don't understand how the code is formatted and what features of the programming language are allows/recommended by the project maintainers.

Definately but i am unsuited for that, as i am barely scratching the understanding of whats possible :)

jklap:
>  bad on the "reading experience" of the developers/users

Except this is what most developers are already used to doing and actually expect :)

> Adding and maintaining the log fields in your class is also error-prone because it can quite easy happen that you use the logger belonging to another class.

Hence why we use tools like SonarQube/Cloud as this type of mistake is flagged as an error

> In my opinion the codebase is very complex as it is, especially parts like the calculation of the revenue of the companies during the operation rounds

Yes that is true, there are parts that are complex yet there are large parts of the code base that are not and provide a low bar to entry that a new developer can work in. Adding Lombok would add "complexity" across the whole code base hence making that bar higher for entry everywhere.

> really useful if we define a set of features that we want to use in the codebase and a formatting that needs to be applied

Completely agree with this-- standardized framework choices, code formatting, etc make it much easier for code to be maintained and for onboarding. I've created several "guides" such as this in the past for working with outside vendors/contractors such that I can pull some stuff together but it would only be a bit of a template and would likely need adjustments.

madoar:
I have removed the usage of lombok from this PR. This PR now only contains normal code refactoring.

neutronc:
Thank you

jklap:
There is some funkiness in StockRound-- trying to merge into my tree and git is complaining and looking at the current file has stuff like this in it?
```
   /**
     * Start the Stock Round. <p>
     * Please note: subclasses that are NOT real stock rounds should NOT call this method
     * (or set raiseIfSoldOut to false after calling this method).
     */
    // called by:
    // GameManager: startStockRound
    // StockRound 1837, 18EU: (start)

    // overridden by:
    // StockRound 1837, 18EU
    // NationalFormationRound, PrussianFormationRound
    public void start() {
```

@madoar can you please review the changes to this file and 1) validate them and 2) fix up anything like the above as it almost looks like a merge failure happened

madoar:
I think this is just historical, see:

https://github.com/Rails-18xx/Rails/blob/50672c4fbee8fb57160a5a34c8a30fab986d4580/src/main/java/net/sf/rails/game/financial/StockRound.java#L103-L114

PR #134, opened 2020-04-10
change message during 18EU minor auction bid/pass to indicate decline…

jklap:
…d to bid

Tweak messaging to indicate that a user declined to bid vs an actual pass


src/test/resources/data/real/18EU_A.report
```diff
@@ -82,109 +82,109 @@ BuysItemFor,Martin,15,105
...
 BuysItemFor,Arne,8,90
 Floats,8
  
 SelectForAuctioning,Martin,10
-PASSES,Arne
-PASSES,Jupp
-PASSES,Volker
+DeclinedToBid,Arne
+DeclinedToBid,Jupp
+DeclinedToBid,Volker
```
madoar, line 97: Two things:
1.  Isn't this more than a message change, but a change to the logic?
2.  I think the value should be `DECLINED_TO_BID`

jklap: Well, I wouldn't consider this a change to the logic except in the fact that it moves the `ReportBuffer.add()` call to under the switch for the two action types. It doesn't actually affect any game play logic

jklap: There is a bit of inconsistency in the value name, ie `DECLINED_TO_BID` vs `DeclinedToBid`-- we have both variants in the code base so I picked the one was similar to the next closest, ie `DeclineToBid`. Would be nice to have this standardized but we have quite a bit a variance and I'd rather spend my time making substantial changes :)

madoar:
I will let @neutronc take a look at this and then let him merge because I don't really understand the implications of changing the label :)

madoar:
@jklap there is a merge conflict in this PR

jklap:
Blah, it had a clean merge!

PR #138, opened 2020-04-10
Replace Maven with Gradle

madoar:


build.gradle
```diff
@@ -4,22 +4,10 @@ import java.text.SimpleDateFormat
 
 plugins {
   id 'java'
-  id 'maven-publish'
   // https://github.com/johnrengelman/shadow
   id 'com.github.johnrengelman.shadow' version '5.2.0'
-  // plugin to check for dependency updates via task dependencyUpdates
-  // https://github.com/ben-manes/gradle-versions-plugin
-  id 'com.github.ben-manes.versions' version '0.28.0'
```
jklap, line 12: I think we should leave this in even with Dependabot as this will also report on plugin versions and gradle versions where as Dependabot only reports on dependencies themselves

build.gradle
```diff
@@ -29,42 +17,20 @@ plugins {
...
     mavenCentral()
-//    maven {
-//        url = 'https://jitpack.io'
-//    }
 }
 
 // Rails variables
-//group = 'net.sf'
-version = '2.2-alpha'
-description = 'Rails (18xx)'
```
jklap, line 40: This is a gradle property so we should add it back (and maybe use it instead of the `railsDescription` we have below?

build.gradle
```diff
@@ -105,7 +70,7 @@ assemble.dependsOn(shadowJar)
 jar {
     manifest {
         attributes(
-                'Main-Class': "${railsMainClass}"
+                'Main-Class': "net.sf.rails.util.RunGame"
```
jklap, line 73: With this can we ditch the railsMainClass defined above on line 26?

build.gradle
```diff
@@ -151,66 +116,3 @@ gitProperties {
...
-//    def rejected = ['alpha', 'beta', 'rc', 'cr', 'm'].any { qualifier -> version ==~ /(?i).*[.-]${qualifier}[.\d-]*/ }
-//    return rejected
-}
-
-dependencyUpdates {
-  gradleReleaseChannel = 'current'
-  rejectVersionIf {
-    isNonStable(it.candidate.version)
-  }
-}
```
jklap, line 215: This is part of the ben-manes plugin so think we should retain it.

build.gradle
```diff
@@ -29,42 +17,20 @@ plugins {
 repositories {
     mavenLocal()
     mavenCentral()
-//    maven {
-//        url = 'https://jitpack.io'
```
jklap, line 33: FYI-- I don't know what this was used for in the Maven build so I just brought it over-- we should confirm that it's not needed anymore. @blentz @neutronc ?

neutronc: Dont know may come from some automatism from eclipse

build.gradle
neutronc: yes

build.gradle
neutronc: nods at JKlap : yes

build.gradle
neutronc: I also think it should stay in there

build.gradle
madoar: `jitpack` is a package repository that allows including projects that are hosted on github as a dependency. As far as I know we don't use a dependency that is only hosted on github. Therefore I think we can remove this.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Execute Tests
+        uses: eskatos/gradle-command-action@v1
+        with:
+          gradle-version: 6.3
+          arguments: clean test
+      - name: Build jar
+        uses: eskatos/gradle-command-action@v1
+        with:
+          gradle-version: 6.3
+          arguments: clean build
```
madoar, line 44: We have two choices how to use `gradle` with GitHub actions:
1.  we use `github-command-action`
2.  we use our wrapper `./gradlew`

The first option is follows the aspired style for GitHub Actions. The second option allows us to specify the Gradle version in a single place instead of in two.

@neutronc @jklap which option do you think is better?

jklap: I'd lean towards the second option-- if we expect developers to use the wrapper then having GitHub actions use it also makes sure it all works

build.gradle
```diff
@@ -102,28 +93,80 @@ wrapper {
 }
 
 // BuildDate function adapted from http://stackoverflow.com/questions/19172565/how-append-date-build-to-versionnamesuffix-on-gradle
-def buildDate() {
+def buildDate = {
```
jklap, line 96: Not saying this is wrong but why the change from a method to a closure?

build.gradle
```diff
@@ -102,28 +93,80 @@ wrapper {
...
+            '--description', stringify(description),
+            '--vendor', stringify(vendor),
+            '--license-file', licenceFile,
             '--icon', './native/rails.png',
             '--file-associations', './native/rails-linux-association.properties',
-            '--app-version', '2.2.0',
+            '--app-version', appVersion,
             '--input', 'build/libs',
-            '--main-jar', 'rails-2.2-beta-all.jar',
+            '--main-jar', "rails-${version}-all.jar",
```
jklap, line 133: You should also be able to get the jar name via shadow and hence don't need to construct the name here-- little bit safer and less error prone?

build.gradle
```diff
@@ -102,28 +93,80 @@ wrapper {
...
+            .replaceAll("release", "0")
+}
+
+task buildDebPackage(type: Exec) {
     commandLine 'jpackage',
             '--verbose',
-            '--name', 'Rails (18xx)',
-            '--description', '"Rails is a moderator program for 18xx games"',
+            '--type', 'deb',
+            '--name', stringify(appName),
```
jklap, line 125: The groovy/gradle way would be to specify this a GString, hence "${appName}"

build.gradle
madoar: Because I think it is semantically more correct. In my opinion we use `buildDate` as a value and not as a method.

build.gradle
madoar: How do you get the filename from shadow?

build.gradle
madoar: This automatically adds `"` around the string?

jklap: Not sure it would be any different than others like your main-jar declaration a couple of lines above?

madoar: The issue with these strings is that they can contain spaces. Because we are using the strings as arguments in a command line we need to wrap them in `"` to tell the shell that it is a single argument and not a list of arguments.

jklap: Yes, understand the reasoning, just don't know how `commandLine` is dealing with arguments as I've never needed to call out to an external command like this.

FYI, I suspect this will fail on the Mac as jpackage is not exported to /usr/bin/ like java is. jpackage only exists in $JAVA_HOME/bin/jpackage

build.gradle
jklap: I believe the shadow plugin replaces the default output jar name with the name of the uber jar

madoar:
Another open question is how do we know the name of the created installed file?
We need to know the filename to perform perform the upload to the `GitHub Releases` page.

@neutronc @jklap do you have any ideas?

build.gradle
madoar: It seems like it is working. I think the reason is that `jpackage` is registered in the system `PATH` variable

jklap: `> A problem occurred starting process 'command 'jpackage''`

This is on a Mac, using `gradle buildDmgPackage` using JDK 14

jklap:
@madoar did you see this jpackage option?
```
  @<filename>
          Read options and/or mode from a file
          This option can be used multiple times.
```

Could be useful in reducing redundancy between the different builds easily.

Regarding the output, can we define it using these?
```
  --name -n <name>
          Name of the application and/or package
  --dest -d <destination path>
          Path where generated output file is placed
          Defaults to the current working directory.
          (absolute path or relative to the current directory)
```

madoar:
> Could be useful in reducing redundancy between the different builds easily.

Let us think about this option in a later PR. For now I think it is ok to have a basic working version that is able to generate native installers for a set of operating systems.

Later on we can refactor this to be more flexible and to allow more code reuse.

> Regarding the output, can we define it using these?
> ```
> --name -n <name>
>          Name of the application and/or package
> --dest -d <destination path>
>          Path where generated output file is placed
>          Defaults to the current working directory.
>          (absolute path or relative to the current directory)
>   ```

I don't think so. We are already using `--name`. And `--dest` is used to specify a folder and not a file. I'm currently thinking about writing a gradle task that searches for a generated executable (detected by its file extension) and then renames it to a fixed name. This fixed name will then be something like `rails-18xx.deb`. The disadvantage of this approach is that the version number gets lost in the filename.

build.gradle
madoar: I think this is a configuration issue on your side. On the Mac VM using GitHub Actions it is working, the folder contains the following files after calling `./gradlew buildDmgPackage`:

```
total 473408
-rw-r--r--   1 runner  staff    26K Apr 10 22:11 1835_20200301_2250_Hakko.rails
-rw-r--r--   1 runner  staff   8.4K Apr 10 22:11 1837_20181226_1653_a.rails
-rw-r--r--   1 runner  staff    13K Apr 10 22:11 1837_20181226_1657_a.rails
-rw-r--r--   1 runner  staff   136M Apr 10 22:15 18xx.log
-rwxr-xr-x   1 runner  staff    18K Apr 10 22:11 LICENSE
-rwxr-xr-x   1 runner  staff   952B Apr 10 22:11 README.md
drwxr-xr-x  10 runner  staff   320B Apr 10 22:14 build
-rw-r--r--   1 runner  staff   5.3K Apr 10 22:11 build.gradle
drwxr-xr-x   3 runner  staff    96B Apr 10 22:11 gradle
-rw-r--r--   1 runner  staff     0B Apr 10 22:11 gradle.properties
-rw-r--r--   1 runner  staff   932B Apr 10 22:11 gradle.yml
-rwxr-xr-x   1 runner  staff   5.6K Apr 10 22:11 gradlew
-rw-r--r--   1 runner  staff   2.9K Apr 10 22:11 gradlew.bat
drwxr-xr-x   7 runner  staff   224B Apr 10 22:11 native
-rw-r--r--   1 runner  staff    16K Apr 10 22:11 pom.xml
-rw-r--r--@  1 runner  staff    87M Apr 10 22:18 rails-18xx-2.2.0.dmg
-rw-r--r--   1 runner  staff    26B Apr 10 22:11 settings.gradle
drwxr-xr-x   6 runner  staff   192B Apr 10 22:11 src
```

You can see that a `rails-18xx-2.2.0.dmg` file has been created. I think most likely `jpackage` is not added to your system path.

jklap:
Wouldn't the output of the files always exist in the same directory and with a known format filename? I would think this is something we can just calculate/hardcode for the time being?

build.gradle
jklap: I'm pretty sure this is not an issue specific to my Mac-- JDK installs on Macs are consistent in location and jpackage is *not* part of the default path. I can't speak to the Mac VM running under GitHub actions but I suspect that that may be a custom JDK install.

madoar: How have you installed JDK 14? Have you done any additional configuration changes?
I know that switching between different JDK versions can be an issue on both Windows and Linux because you need to watch out that all paths (`java`, `javac`, `JAVA_HOME`, etc.) are changed correctly. What output do you get if you fetch the version of a more obscure application bundled with the JDK (for example `jjs -version`)?

Independent I don't think I can help you here. The issue should be fixed if you add `jpackage` to your system or user path, then it should be detectable when you run `jpackage` in your terminal.

jklap: Yes. No. Switching versions on a Mac is simple and yes, I'm currently using JDK 14 for this test.

Applications that are bundled with JRE/JDKs on a Mac have it built into the application-- the JRE/JDK is not exposed.

Adding jpackage to a user's PATH is not an appropriate solution-- the user should be able to check out the code and build via ./gradlew without needing to do anything special, other than obviously making sure an appropriate version of the JDK is installed-- which on a Mac is managed by the packages released by Oracle or OpenJDK or Adopt and by default is exposed on the command line without any additional effort by the user.

The plugin that I was playing with was able to find jpackage without an issue so obviously it's right where it's expected to be even if it's not on my PATH :) This is the problem we have with switching to calling out to command line tools directly :( As mentioned earlier, we should really be making the call out relative to JAVA_HOME as that then removes any dependencies on a user's PATH AND given that JAVA_HOME is a well-known env var shouldn't run into any issues as a user that doesn't have JAVA_HOME set is going to have much bigger problems doing anything with Java.

madoar: I had my own experiences with `JAVA_HOME` set wrong or not set at all. I wouldn't expect this to be set correctly on all systems. Also we are talking about a developer setup here. If we add to our documentation that `jpackage` needs to be set to build the installers locally I think this is fine. The normal end-user normally only downloads an installer from the vendor of the software. In our case the user should download the installer from https://github.com/Rails-18xx/Rails/releases

madoar: Independent of the discussion:

This is currently only the first version of the `jpackage` integration. While it may not be perfect it does its job, namely it allows the creation of a native installer for all major platforms.

In the future we can extend this in multiple directions if we want to:
-  change to the `jpackage` gradle plugin if we can get it working
-  add some extra checks to our current tasks. Possible checks are for example:
   -  checking for a compatible operating system (e.g. `deb` can only be built on a linux system)
   -  checking/searching for a `jpackage` installation

In my opinion we don't need to fix this issue in this PR.

Also, please be aware that I can't test the scripts for Windows and MacOS outside `GitHub Actions`, where it is apparently working.

jklap: On Macs the JDK version is managed by the OS, including setting JAVA_HOME. So yes, I would expect this to be set correctly on all Macs. And yes, we are talking about a developer setup here. 

I still disagree, we should fix the build so that a user does not need to do anything out of the ordinary. Needing a user to do something other than make sure JAVA_HOME is set correctly is out-of-the-norm with development projects in this day and age.

madoar: Let us take two steps back, I don't think I can follow you.

How are you managing your JDK installations? According to https://stackoverflow.com/questions/52524112/how-do-i-install-java-on-mac-osx-allowing-version-switching there are a lot of options to manage your JDK installations. Depending on your choices I guess you can face different difficulties.

Which applications are added to your path when you switch your JDK? I expect at least `java` and `javac` are added, otherwise quite a lot of things are difficult to do.

What exactly do you expect me to do? I don't own a mac. This means that every change I do results in me stumbling in the dark. It makes more sense if you add the necessary parts to the `dmg` creation task, compared to if I try to add these parts.

jklap: The "official" JDK path-- either with Oracle's package or OpenJDK (or Adopt). Any other solution as listed on that page such as brew, MacPorts, etc would be out of scope of our support.

My suggestion is to make ALL the jpackage executions work based off of JAVA_HOME. If you get that working locally for your env in Linux, given that JAVA_HOME is pretty much a requirement for any dev env, then it should work fine for Mac, for Windows. Nothing super complex or tricky.

madoar: @jklap can you try with the current version again please? It makes use of `JAVA_HOME` to try to detect `jpackage`

build.gradle
madoar: I have readded the `ben-manes` plugin

build.gradle
madoar: I have readded the `ben-manes` plugin

madoar:
Ok I think this PR is now ready to be merged.

@jklap I have used the `JAVA_HOME` environment variable to resolve the path to the `jpackage` executable. This should hopefully solve your issue.

GitHub Actions is now setup to publish a new pre-release whenever a push has been made to the master branch.

build.gradle
```diff
@@ -142,75 +109,118 @@ wrapper {
...
+            '--icon', './native/rails.ico',
+            '--file-associations', './native/rails-osx-association.properties',
+            '--app-version', appVersion(),
+            '--input', 'build/libs',
+            '--main-jar', shadowJar.archiveName,
+            '--mac-package-name', 'rails-18xx'
+}
+
+task renameExecutables(type: Copy) {
+    from('.') {
```
jklap, line 218: Would suggest we instead set a destination path in the jpackage builds (maybe build/jpackage or something in build) and then limit this `from` to that directory.

madoar: Let us do that in a later PR

PR #139, opened 2020-04-10
Add option to hide networth and OR during game

Jungorend:
Some friends and I play play 18xx with one Rails app over a zoom, but prefer that none of us can know the net worth mid-match. This adds the option to disable these views, but leaves the default configuration as it was before.


jklap:
Note that this change only affects the GameStatus window-- the Report window will still contain details on networth and OR value changes.

Jungorend:
That's fine, the report window can optionally be closed during the game so the impact is minimal in my experience. I can modify this too if preferred.

The one thing I have noticed is by hiding these elements, as the GameStatus Window is made smaller, the available trains can get scrunched. If preferred I will keep the Networth/OR UI fields, and just blank them out to retain consistency.

jklap:
Actually I like that suggestion-- maybe "***" them out or something to show they are actually hidden vs missing. Plus leaving the fields in place make the UI/UX stuff easier to test across all of the games as it's not messing with the components, just the actual content of a couple of them

jklap:
I was thinking about this last night and it seems to me this should really be a game option and not a player option, similar in nature to the tile highlight and revenue calculation suggest options in game setup. And in that context I think we should look at what it would take to also remove this detail from the report window. While you are currently playing with a single instance over Zoom that is likely one of the less use-cases with PBEM and real-time over a shared folders more frequently used. If revenue calc suggestions are a game-level option then I think this should be to to keep a level playing field?

madoar:
> I was thinking about this last night and it seems to me this should really be a game option and not a player option, similar in nature to the tile highlight and revenue calculation suggest options in game setup.

What do you mean by player option vs. game option?

> While you are currently playing with a single instance over Zoom that is likely one of the less use-cases with PBEM and real-time over a shared folders more frequently used.

I'm also normally playing with friends by screen sharing.

jklap:
Game options are set during the game creation-- and are game specific (ie "Unlimited D trains") or game play specific (ie "Revenue calculation: suggest") and are stored in the game file affecting all players, ie:

Screen Shot 2020-04-11 at 9 00 43 AM

Player options are specific to the player's config, locally such as sound preferences, train colors, ie:

Screen Shot 2020-04-11 at 9 00 54 AM

My comment regarding use-case playing styles (PBEM, screen-share, etc) is that this current change only really works for screen-share as the other styles would require you to make sure all other players running Rails locally set the right options for, what seems like, a game related option-- ie limit display so all have the same knowledge...

madoar:
I agree this kind of setting should be saved in the game save file

Jungorend:
I've updated the PR to be a game configuration option for 1830. If any other games want to have this option, all they should need is to add a line like:
`<GameOption name="NetworthHidden" type="toggle" default="no" />`
to their GameOptions.xml.

madoar:
> I've updated the PR to be a game configuration option for 1830. If any other games want to have this option, all they should need is to add a line like:
> `<GameOption name="NetworthHidden" type="toggle" default="no" />`
> to their GameOptions.xml.

I think we should add this to all game variants.

PR #142, opened 2020-04-10
code cleanup

jklap:
Misc code cleanup

### DIFF
--- a/src/main/java/net/sf/rails/game/GameManager.java
+++ b/src/main/java/net/sf/rails/game/GameManager.java
@@ -115,9 +115,7 @@ public class GameManager extends RailsManager implements Configurable, Owner {
      */
     protected GuiHints guiHints;
 
-    /**
-     * Flags to be passed to the UI, aiding the layout definition
-     */
+    /** Flags to be passed to the UI, aiding the layout definition */
     protected final EnumMap<GuiDef.Parm, Boolean> guiParameters = new EnumMap<>(GuiDef.Parm.class);
 
     protected GenericState<StartPacket> startPacket = new GenericState<>(this, "startPacket");
@@ -694,7 +692,6 @@ public class GameManager extends RailsManager implements Configurable, Owner {
         if (action instanceof NullAction && ((NullAction) action).getMode() == NullAction.Mode.START_GAME) {
             // Skip processing at game start after Load.
             // We're only here to create PossibleActions.
-            result = true;
             startGameAction = true;
         } else if (action != null) {
             // Should never be null.
@@ -763,7 +760,7 @@ public class GameManager extends RailsManager implements Configurable, Owner {
 
         // logging of game actions activated
         for (PossibleAction pa : possibleActions.getList()) {
-            log.debug("{} may: {}", getCurrentPlayer().getId(), pa.toString());
+            log.debug("{}", pa.toString());
         }
 
         return result;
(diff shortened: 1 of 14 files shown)


src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -115,9 +115,7 @@
      */
     protected GuiHints guiHints;
 
-    /**
-     * Flags to be passed to the UI, aiding the layout definition
-     */
+    /** Flags to be passed to the UI, aiding the layout definition */
```
madoar, line 118: We should decide on a code style here. I normally use multi-line comments for class definitions, class fields and class methods. Only inside methods I use single-line comments.

What about you @jklap @neutronc?

src/main/java/net/sf/rails/ui/swing/GameSetupController.java
```diff
@@ -219,14 +214,10 @@ public void actionPerformed(ActionEvent arg0) {
...
-                    return ComparisonChain.start()
-                            .compare(b.lastModified(), a.lastModified())
-                            .compare(a.getName(), b.getName())
-                            .result();
-                }
-            });
+            SortedSet<File> recentFiles = new TreeSet<>((a, b) -> ComparisonChain.start()
+                    .compare(b.lastModified(), a.lastModified())
+                    .compare(a.getName(), b.getName())
+                    .result());
```
madoar, line 220: Do we want to use `ComparisonChain` or `Comparator.comparing` with `Comparator.thenComparing`?

src/main/java/net/sf/rails/game/GameManager.java
jklap: I tend to use multiple for class and methods, single line for inside methods.

For class fields I tend to also stick with single line unless it's long enough where it makes the comment too long. Multiple line for class fields just seems to take up a lot of space and doesn't really add value unless class and method where you should have parameters, return details, class description etc.

src/main/java/net/sf/rails/ui/swing/GameSetupController.java
jklap: Doesn't matter to me-- didn't write the original code using `ComparisonChain` so don't know if there is a preference

madoar: I didn't see that this call was there before this PR. In this case I think it is fine if we let it remain as is. On the other hand I would like to see if it also works with build in Java features (i.e `Comparator.comparing` and `Comparator.thenComparing`) :)

src/main/java/net/sf/rails/game/GameManager.java
neutronc: agreed

madoar: When do you use:

```
// ...
```

compared to:

```
/** ... */
```

?

jklap: Class and methods only. Even multiline comments inside of methods I use `//` as `/* */` is just plain harder to scan/read especially on a single line. The benefit of `/* */` is when you have that multiline comment as you don't need to preface every line with a `//` BUT in those cases using the opening `/*` and the closing `*/` tend to be on their OWN lines to increase the readability of the comment. In a method this just adds extra lines that get in the way (vs the top of a method or class where it is not that important because you aren't trying to follow code flow.

PR #143, opened 2020-04-10
return failure on file compare issues

jklap:
Fixes the return value when reload/auto-load fails during the action compare


madoar:
Should the following line return `false` as well?

https://github.com/Rails-18xx/Rails/blob/16cdd9b7d3c74bbd4d05ec31e3da6d1d49c68b6d/src/main/java/net/sf/rails/game/GameManager.java#L1027

jklap:
Yep, probably should

PR #144, opened 2020-04-11
bug fix on reload executed actions size calc

jklap:
Fixed a bug during a reload/auto-load where we reference the size of the executed actions array while we are actively adding items where as we really wanted the initial size

### DIFF
--- a/src/main/java/net/sf/rails/game/GameManager.java
+++ b/src/main/java/net/sf/rails/game/GameManager.java
@@ -982,11 +982,12 @@ public class GameManager extends RailsManager implements Configurable, Owner {
 
         // Check action identity
         int index = 0;
-        int executedCnt = executedActions.size(); // save off the current # of executed actions as it will grow as we execute newly loaded
+        // save off the current # of executed actions as it will grow as we execute newly loaded
+        int executedActionsCount = executedActions.size();
         PossibleAction executedAction;
         try {
             for (PossibleAction savedAction : savedActions) {
-                if (index < executedCnt) {
+                if (index < executedActionsCount) {
                     executedAction = executedActions.get(index);
                     if (!savedAction.equalsAsAction(executedAction)) {
                         DisplayBuffer.add(this, LocalText.getText("LoadFailed",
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -982,10 +982,11 @@ protected boolean reload(GameAction reloadAction) {
 
         // Check action identity
         int index = 0;
+        int executedCnt = executedActions.size(); // save off the current # of executed actions as it will grow as we execute newly loaded
```
madoar, line 985: Please add the comment above the this line. This will prevent formatters from cutting it up in multiple lines.

madoar, line 985: I prefer writing out words instead of shortening them:

```
int executedCount = ...
```

jklap: Agree, except when there are commonly well known abbreviations such as Cnt, Idx AND those abbreviations are secondary words

jklap: Correctly configured formatters should never chop down comments :) But will move the comment above

madoar: We are talking about 2 letters that you save by using an abbreviation. I can understand when you want to shorten an 8 letter word to 3 letters, but reducing 5 letters to 3 does not help readability in my opinion.

madoar: I think this is exactly what happens if you apply the IntelliJ or Eclipse formatter with default settings. I think it is quite likely that someone (like me) applies such a formatter on the code

neutronc: Comments on separate lines would be my preference as well sorry :)

neutronc: PI also vote for the 5Letter version instead of Three :)

jklap:
Sorry guys but I am refusing, absolutely, to rename the branch from pr/exec_action_cnt to pr/executed_action_count.

Meanwhile the other changes you've requested have been made ;)

PR #145, opened 2020-04-11
recent files & loading of last_rails support

jklap:
This PR consists of several changes:
* Support to "load" a last_rails file from the "Load Game" option (in addition obviously to an actual saved game file)
* Converted the "last_rails" strings to a single static
* Enable "filtering" in the Load Game file dialog to our default file extensions along with any user defined saved file extension vs allowing a user to select just any file

It also contains changes to the "Load Recent" option that adds three configuration options to:
* Display only last_rails files (boolean)
* Support a user defined include regex
* Support a user defined exclude regex

The use-case for the filtering is to support reducing the # of files found & displayed. My default save location, in the directory hierarchy, has almost 3K game files in it. While playing we tend to move older game files out of the way into an archive directory yet these files are still "found" by the current logic. The include/exclude allow me to define filters so that the Load Recent is actually useful by excluding these archive directories. The use of the last_rails boolean allow me to filter out all actual game files so I don't have to "hunt down" the most recent saved file for a game, I can instead just always select the last_rails file.


src/main/java/net/sf/rails/ui/swing/GameSetupController.java
```diff
@@ -183,25 +189,44 @@ private void startNewGame() {
...
+            return false;
+        }
+        switch (ext) {
+            case GameUIManager.DEFAULT_SAVE_EXTENSION:
+            case GameUIManager.DEFAULT_SAVE_POLLING_EXTENSION:
+                return true;
+            default:
+                return ext.equals(savedFileExtension);
+        }
+    }
```
madoar, line 205: I can't identify what this method does by reading the name. Is there maybe a more fitting method name? Alternatively we should add a comment to the method.

src/main/java/net/sf/rails/util/GameLoader.java
```diff
@@ -42,6 +46,20 @@ public static void loadAndStartGame(File gameFile) {
...
+        if ( GameUIManager.DEFAULT_SAVE_POLLING_EXTENSION.equals(StringUtils.substringAfterLast(gameFile.getName(), ".")) ) {
+            // read the filename from the last rails file
+            log.debug("loading current game file from last_rails {}", gameFile);
+            try {
+                String gameFileStr = FileUtils.readFileToString(gameFile, StandardCharsets.ISO_8859_1).trim();
+                gameFile = new File(gameFile.getParentFile(), gameFileStr);
+            }
+            catch (IOException e) {
+                log.warn("unable to load {}", gameFile);
+                return;
```
madoar, line 59: What does the application do in this case?

jklap: Not my code-- haven't reviewed

src/main/java/net/sf/rails/ui/swing/GameSetupController.java
jklap: Method name + argument is defining-- is Our file?

madoar: Yes but how do you define `our file`?
If you mean files belonging to Rails then wouldn't `belongsToRails(...)` be more appropriate?

PR #147, opened 2020-04-12
Repository Cleanup

madoar:
This PR:
- removes the now unnecessary pom.xml file
- removes the old travis link from README


jklap:
Looks fine :)

neutronc:
I am still having problems with Eclipse not picking up the gradle config and thus i am not really able to build anything without maven ? Anyone has some insight ?...

PR #148, opened 2020-04-12
Improve the Grade Pipeline

madoar:
This PR:
- sets a valid tag name
- sets a better output folder for the generated installers


jklap:
|`build.gradle:35`|
|---|
Shouldn't this be under `build/`?

madoar:
By this you mean the `output` field, right? It depends. The `build` directory is mainly managed by Gradle. This means that it contains a number of other files and folders. If we don't want to add our installers to the root project folder (i.e. `.`) then I think we should add it to a folder which is fully controlled by us therefore the new folder.

As a sidenote:
This PR still writes the final installers, i.e. `rails-18xx.(deb|rpm|exe|dmg)` to the root project folder because I don't know how the upload action in the pipeline behaves. If we are unlucky it uploads the installer including the folder it resides in, which I don't want.

jklap:
Given we are creating the artifacts in a gradle build I think `build/` is a completely appropriate place to store them, just like the use of `target/` in Maven builds.

madoar:
How about we use `build/installers` as a compromise? It adds the installers to the `build` folder (where they should go) and still gives us full control because of the new subdirectory.

jklap:
Oh totally-- yes, sorry, did not mean to imply it should be at `build/`' but just somewhere inside of that directory structure!

madoar:
I have changed the output directory to `./build/installers`

PR #149, opened 2020-04-12
Refactor the CompanyManager class

madoar:
This PR:
- adds the `18xx_test.log` file to gitignore
- refactors the `CompanyManager` class

### DIFF
--- a/src/main/java/net/sf/rails/game/CompanyManager.java
+++ b/src/main/java/net/sf/rails/game/CompanyManager.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import net.sf.rails.common.LocalText;
 import net.sf.rails.common.parser.Configurable;
@@ -13,8 +14,11 @@ import net.sf.rails.common.parser.Tag;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.util.function.Predicate.not;
 
 public class CompanyManager extends RailsManager implements Configurable {
+
+    private static final Logger log = LoggerFactory.getLogger(CompanyManager.class);
 
     /**
      * This is the name by which the CompanyManager should be registered with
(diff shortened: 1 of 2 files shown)


madoar:
> Can you please back out the whole file code (re)formatting?

What specific changes do you mean? I'm fine with to withdraw the accidental comment formatting changes. The other changes are intended to be part of this PR and a lot of them were even done manually (like removing unnecessary linebreaks).

> 1) we can't see the actual logic/code changes you've made from syntactic changes and hence makes it shy of impossible to actual review the real changes

I did not add any kind of logical change. It is all syntactic in nature (except some added `final` keywords).

> 2) if we are going to start making whole file reformats we should really put a pause on other changes as this level of change will make it impossible for a user to merge their PRs.

I don't see a problem here. In my opinion it is totally acceptable that code changes lead to merge conflicts in other PRs. Not all change are compatible with each other.

If this is an issue then we need to force the users to follow a uniform set of codestyle guidelines. This means we need to add these as blockers for our PRs. [Codacy](https://www.codacy.com/) for example allows this.

src/main/java/net/sf/rails/game/CompanyManager.java
```diff
@@ -123,54 +132,53 @@ public void configureFromXML(Tag tag) throws ConfigurationException {
...
 
+                    mPrivateCompanies.put(name, privateCompany);
+                    lPrivateCompanies.add(privateCompany);
                 } else if (company instanceof PublicCompany) {
-                    ((PublicCompany)company).setIndex (numberOfPublicCompanies++);
-                    mPublicCompanies.put(name, (PublicCompany) company);
-                    lPublicCompanies.add((PublicCompany) company);
+                    final PublicCompany publicCompany = (PublicCompany) company;
+
+                    publicCompany.setIndex(numberOfPublicCompanies++);
```
neutronc, line 164: Hm where is that defined earlier ?

jklap:
I'd consider changes such as those in `considerAllPrivates()` a logic change. While yes, it supposedly has the same result is accomplishes it in another way (in this case `for` to `stream`) hence is a logic change.

The removal of line breaks I'm totally for and tend ("tend") to not lead to merge failures. In the case of this PR I'd say we keep the line break changes but stop further (including me!) pending discussion(s) as raised in Slack.

>> if we are going to start making whole file reformats we should really put a pause on other changes as this level of change will make it impossible for a user to merge their PRs.

> I don't see a problem here. In my opinion it is totally acceptable that code changes lead to merge conflicts in other PRs. Not all change are compatible with each other.

Totally acceptable certainly, but for something as minor as a code (re)format, less acceptable given that we can *avoid* the issue with failed merges and having to manually merge for something this trivial. And yes, obviously some changes are not compatible with each other but again, we are talking syntactical here-- and these are type of merge failures that we can avoid.

> If this is an issue then we need to force the users to follow a uniform set of codestyle guidelines. This means we need to add these as blockers for our PRs. Codacy for example allows this.

Agreed-- hence why I spun up an instance of SonarCloud (which tends to be considered the best-of-breed for Java applications). This also gets us not only the ability to define code style rules but also benefits of sec and static analysis. Some of my early commits were actually code changes based on initial review of low-hanging fruit from running my own instance of SonarQube-- things it caught as outright bugs such as equality string compares vs object, etc. But I think we have a ways to go before we can start to institute this as a PR gate. Would love to get to that point, just think we have some work to do first

PR #151, opened 2020-04-13
Deprecated fix

sknop:
The tiniest fix, just to test out the hard work you guys put into the gradle port.


src/main/java/net/sf/rails/game/TrackConfig.java
```diff
@@ -132,6 +132,7 @@ public String toString() {
      * a description of the connections
      * @deprecated Use {@link #getConnectionString(MapHex,Tile,HexSide,Station)} instead
      */
+    @Deprecated
```
jklap, line 135: LOL, I was wondering how long it would take for someone to fix this. Evidently not long!

PR #153, opened 2020-04-13
improved customizable message for Slack and Discord and removed singl…

jklap:
…eton

Improved support for customizing the message such as:
* Your turn @Jason says Randy, StockRound 1 of 1830

Also removed the singleton implementation

### DIFF
--- a/src/main/java/net/sf/rails/common/notify/Discord.java
+++ b/src/main/java/net/sf/rails/common/notify/Discord.java
@@ -7,6 +7,8 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.text.StrSubstitutor;
+import org.apache.commons.text.StringSubstitutor;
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/rails/common/notify/Discord.java
```diff
@@ -84,18 +90,21 @@ public void update(String text) {
         public Observable getObservable() {
             return pm.getCurrentPlayerModel();
         }
-    }
 
-    public static void notifyOfGameInit(final RailsRoot root) {
-        instance.init(root);
+        public Player getFormerPlayer() {
+            return formerCurrentPlayer;
```
neutronc, line 95: hm formerCurrentPlayer (sounds logically but on the other side horrific :))

jklap: Haha, yes, yes it is quite ugly. And to be sure, it is not my invention, see SoundEventInterpreter :)

PR #154, opened 2020-04-13
Use release tagging to gate releases.

blentz:
This PR moves code releases to a manual process, triggered by pushing a tag into the repo.

The tag format is `releases/v*`, so it should be something like `releases/v2.2.0`


.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+        run: ./gradlew buildRpmPackage
+      - name: Rename Installers
+        if: github.ref == 'refs/heads/master'
+        run: ./gradlew renameExecutables
+      - name: Upload Deb Installer
+        if: github.ref == 'refs/heads/master'
+        uses: actions/upload-artifact@v1
+        with:
+          name: deb
+          path: rails-18xx.deb
```
madoar, line 54: This can be removes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          java-package: jdk
+      - name: Execute Tests
+        run: ./gradlew clean test
+      - name: Build jar
+        run: ./gradlew clean build
+      - name: Create Exe Installer
+        run: ./gradlew buildExePackage
+      - name: Rename Installer
+        if: github.ref == 'refs/heads/master'
+        run: ./gradlew renameExecutables
```
madoar, line 75: This can be removed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          java-package: jdk
+      - name: Execute Tests
+        run: ./gradlew clean test
+      - name: Build jar
+        run: ./gradlew clean build
+      - name: Create Dmg Installer
+        run: ./gradlew buildDmgPackage
+      - name: Rename Installer
+        if: github.ref == 'refs/heads/master'
+        run: ./gradlew renameExecutables
```
madoar, line 96: This can be removed

madoar:
In addition we can use the version contained in the tag and pass it to `jpackage` and `gradle` as the used version string:
- https://github.community/t5/GitHub-Actions/How-to-get-just-the-tag-name/td-p/31595 (use `${{github.ref}}` instead)
- https://stackoverflow.com/questions/32717251/how-to-set-project-version-by-passing-version-property-on-gradle-command-line

.github/workflows/release.yml
```diff
@@ -44,16 +42,13 @@ jobs:
       - name: Create Rpm Installer
         run: ./gradlew buildRpmPackage
```
madoar, line 43: We should pass the `version` to this command and the `buildDebPackage` command as well, see https://stackoverflow.com/a/40239092

.github/workflows/release.yml
```diff
@@ -77,10 +72,8 @@ jobs:
       - name: Create Exe Installer
         run: ./gradlew buildExePackage
```
madoar, line 73: We should pass the `version` to this command as well, see https://stackoverflow.com/a/40239092

.github/workflows/release.yml
```diff
@@ -104,10 +97,8 @@ jobs:
       - name: Create Dmg Installer
         run: ./gradlew buildDmgPackage
```
madoar, line 98: We should pass the `version` to this command as well, see https://stackoverflow.com/a/40239092

gradle.yml
```diff
@@ -2,9 +2,9 @@ name: Java CI with Gradle
 
 on:
   push:
-    branches: [ rails_2_develop ]
+    branches: [ master ]
```
madoar, line 5: This file can be deleted

PR #155, opened 2020-04-13
new game w/poller enabled should ask to save file

jklap:
Currently when auto-load/save is enabled automatically via user configuration the user still has to:
* Save the game for the first time (especially important if they are not the first player!)

In addition, polling isn't automatically enabled because that game file is saved *after* polling had it's change to initialize.

And finally the UI is not enabled/disabled appropriately if the first user is not the user that started the game (instead the UI is left enabled, ie the "default" state).

This PR addresses all of these issues by:
* Forcing a save of the game after the the UI has completed it's initial initialization
* Enabling polling by moving initialization to after the UI's initialization is complete (and hence it can then also trigger the forced save)
* Set the myTurn variable appropriate so that when the UI is drawn for the first time it is enabled/disabled correctly.

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
+++ b/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
@@ -182,6 +182,15 @@ public class GameUIManager implements DialogOwner {
         for (Player player : getPlayers()) {
             currentGuiPlayerNames.add(player.getId());
         }
+
+        localPlayerName = System.getProperty("local.player.name");
+        if (!Util.hasValue(localPlayerName)) {
+            localPlayerName = Config.get("local.player.name");
+        }
+        if (autoSaveLoadStatus > 0) {
+            myTurn = getCurrentPlayer().getId().equals(localPlayerName);
+            log.debug("starting game with my turn: {}", myTurn);
+        }
     }
 
     private void initWindowSettings() {
@@ -317,13 +326,8 @@ public class GameUIManager implements DialogOwner {
         splashWindow.notifyOfStep(SplashWindow.STEP_INIT_SOUND);
         SoundManager.notifyOfGameInit(railsRoot);
 
-        // TODO: switch to injecting Discord to loosely couple
         new Discord(this, railsRoot);
         new Slack(this, railsRoot);
-
-        if (gameWasLoaded && autoSaveLoadStatus > 0) {
-            startAutoSaveLoadPoller();
-        }
     }
 
     public void startLoadedGame() {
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/rails/ui/swing/GameUIManager.java
```diff
@@ -182,6 +182,15 @@ public void init(RailsRoot root, boolean wasLoaded, SplashWindow splashWindow) {
         for (Player player : getPlayers()) {
             currentGuiPlayerNames.add(player.getId());
         }
+
```
madoar, line 185: What does `System.getProperty("local.player.name")` contain and where is it set?

src/main/java/net/sf/rails/ui/swing/GameUIManager.java
```diff
@@ -182,6 +182,15 @@ public void init(RailsRoot root, boolean wasLoaded, SplashWindow splashWindow) {
...
+        localPlayerName = System.getProperty("local.player.name");
+        if (!Util.hasValue(localPlayerName)) {
+            localPlayerName = Config.get("local.player.name");
+        }
+        if (autoSaveLoadStatus > 0) {
+            myTurn = getCurrentPlayer().getId().equals(localPlayerName);
```
madoar, line 191: I know this is kind of independent to this PR:
Isn't `myTurn` only useful in a game that is played with `discord` etc.?
If you do a "normal" game where all players "share" the same screen the variable should always be `true` right?

src/main/java/net/sf/rails/ui/swing/GameUIManager.java
jklap: I believe it's a fallback allowing a user to set their player name via a system property when starting the game vs configuration. This logic existed previous, not sure if it's value anymore but since I didn't write it and don't know the use case I'm not comfortable removing it

src/main/java/net/sf/rails/ui/swing/GameUIManager.java
jklap: `myTurn` was used (and still is) in the originally polling logic in game manager(s) to determine if/when it should run certain UI updates and calculations.

And yes, if a game is not using the poller then myTurn is always true. This is the reason for this specific check-- we only change `myTurn` if the poller is enabled.

jklap: And my assumption on the mention of Discord here is not a reference to our Discord and Slack notification classes (which are independent of polling or not) and instead to the user using Discord (or any IM communication tool) to coordinate game play?

jklap:
Yes, this can wait!

PR #157, opened 2020-04-13
fix quit issue on Mac

jklap:
If a user uses the OS File menu to quit Rails (vs the menu inside of the Status window) it by-passes the termination logic we have in place (which, for example, saves off current window sizes and locations). This PR fixes that by added a event handler for `Desktop` events (which may also be generated on other platforms FYI.

This also removes the dummy desktop window created (which needs confirmation it does not cause any issues on Linux and Windows as I've confirmed it continues to be fine on the Mac).

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/StatusWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/StatusWindow.java
@@ -321,7 +321,7 @@ public class StatusWindow extends JFrame implements ActionListener, KeyListener,
                 Class.forName(gameStatusClassName).asSubclass(GameStatus.class);
             gameStatus = gameStatusClass.newInstance();
         } catch (Exception e) {
-            log.error("Cannot instantiate class " + gameStatusClassName, e);
+            log.error("Cannot instantiate class {}", gameStatusClassName, e);
             System.exit(1);
         }
 
@@ -385,6 +385,26 @@ public class StatusWindow extends JFrame implements ActionListener, KeyListener,
                 guiMgr.getWindowSettings().set(frame);
             }
         });
+
+        if ( Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.APP_QUIT_HANDLER) ) {
+            Desktop.getDesktop().setQuitHandler((e, r) -> {
+                if ( JOptionPane.showConfirmDialog(frame, LocalText.getText("CLOSE_WINDOW"), LocalText.getText("Select"), JOptionPane.OK_CANCEL_OPTION) == JOptionPane.OK_OPTION ) {
+                    frame.dispose();
+                    guiMgr.terminate();
+                    r.performQuit();
+                }
+                else {
+                    r.cancelQuit();
+                }
+            });
+        }
+
+//        SwingUtilities.invokeLater(() -> {
+//            JFrame main = new JFrame("java.awt.Desktop");
+//            main.setSize(new Dimension(600, 400));
+//            main.setLocationRelativeTo(null);
+//            main.setVisible(true);
+//        });
 
         gameUIManager.packAndApplySizing(this);
     }


jklap:
@madoar @neutronc Can you both please verify this PR does not have an adverse effect on your platforms (Windows & Linux)? The addition of the handler is conditional based on whether the platform supports the event. 

In addition, the removal of the dummy window should be tested to make sure everything looks & works normally.

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
```diff
@@ -386,6 +386,26 @@ public void componentResized(ComponentEvent e) {
             }
         });
 
+        if ( Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.APP_QUIT_HANDLER) ) {
+            Desktop.getDesktop().setQuitHandler((e, r) -> {
+                if ( JOptionPane.showConfirmDialog(frame, LocalText.getText("CLOSE_WINDOW"), LocalText.getText("Select"), JOptionPane.OK_CANCEL_OPTION) == JOptionPane.OK_OPTION ) {
+                    frame.dispose();
+                    guiMgr.terminate();
+                    r.performQuit();
```
madoar, line 394: What is `r`? Can you please give it a more meaningful name?^^

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
```diff
@@ -386,6 +386,26 @@ public void componentResized(ComponentEvent e) {
...
+                }
+            });
+        }
+
+//        SwingUtilities.invokeLater(() -> {
+//            JFrame main = new JFrame("java.awt.Desktop");
+//            main.setSize(new Dimension(600, 400));
+//            main.setLocationRelativeTo(null);
+//            main.setVisible(true);
+//        });
```
madoar, line 407: Can this be removed?

jklap: Did you see my first comment in the PR? I'd prefer not to remove this until you and @neutronc are able to verify this removal doesn't break the UI.

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
jklap: :+1:

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
jklap: I'll nuke it anyway-- if we need to restore that is what we have version control for

jklap:
@madoar I'd like to make your two request changes in an upcoming PR-- I've already got additional changes to this code made (for another feature) and hence making the change in my existing code based is not pretty.

madoar:
I have tested three different ways to exit the application:

1.  click the big `x` in the `Rails: Game Status` window frame -> the application shows the confirm dialog and exits after I confirm
2.  click on `File -> Quit` in the menu bar -> the application directly exits without showing the confirm dialog
3.  right click on the open `Rails: Game Status` window in the task bar and click Exit -> the application shows the confirm dialog and exits after I confirm

jklap:
So #1 and #3 are handled by the window closing handler and the Status window menu handler.

#2 is platform specific. For the Mac this is the logic that handles the Menu bar closing:
```
Desktop.getDesktop().isSupported(Desktop.Action.APP_QUIT_HANDLER)
```

So that means your platform does have the APP_QUIT_HANDLER support. 

That is fine in the context of this PR but we should open an issue to get your platform resolved. The primary implication of this is that the `terminate()` method isn't being called so your window settings aren't getting saved.

jklap:
And @madoar I'm also assuming during your testing you didn't see any issues, the windows all showed up as expect, etc?

madoar:
I think all the windows show up as expected.

jklap:
Thanks @mador for testing!

PR #163, opened 2020-04-14
gradle build script & jpackage cleanup

jklap:
Changes:
* Removed mavenLocal() as an artifact source
* Added check for jpackage exec existence before we attempt to actually use it
* Some GString to String changes
* Moved everything from ./native to ./src/main/resources/build/ and ./src/main/resources/images/icon/
* Removed old packaging files (deb, etc) no longer used with jpackage
* Removed duplicated icon files and changed references to correct files
* Added test icons for saved games & last rails for Mac


jklap:
@madoar why would adding a executable check cause these builds to fail?

jklap:
So further looks--- it's failing in the test jobs-- not sure why the `jpackage()` method would even be called in testing?

jklap:
Duh-- been a while since I crafted new gradle scripts-- the problem here is that a task definition as done here is actually "executed" (ie configured) during initial setup in gradle but not actually executed (ie calling jpackage). The change is to move the command into the actual task execution part of the task definition. Will be pushing the fix in a couple

jklap:
Blah-- that didn't work. Still looking

src/main/resources/build/last_rails-osx-association.properties
```diff
@@ -0,0 +1,4 @@
+extension=last_rails
+mime-type=application/rails-18xx-last
+icon=./src/main/resources/images/icon/last_rails.icns
+description=Rails polling file
```
madoar, line 4: I don't think these files (including the icon file(s)) belong in the `src/main/resources` directory. My reason is that these files are not part of the application sources. Instead they are part of the build configuration. Therefore I put the in another folder on the root level to reflect this.

build.gradle
```diff
@@ -14,34 +14,42 @@ plugins {
...
+    } else if (OperatingSystem.current().isMacOsX()) {
+        jpackageExe = "${System.getenv('JAVA_HOME')}/bin/jpackage"
+    } else {
+        throw new StopActionException("Unknown operating system: ${OperatingSystem.current()}")
+    }
+
+    if ( ! new File(jpackageExe).exists() ) {
+        throw new StopActionException('jpackage not found (JDK 14+) needed to package')
+    }
+    jpackageExe
```
madoar, line 52: I like it!

src/main/resources/build/last_rails-osx-association.properties
jklap: And yet they don't belong at the root either :) I think the assumption is that `src/main/resources` are only application "included" resources which isn't really correct, they can be anything used for the application, even for building the final apps (as often there is filtering/templating involved). I've seen some cases where these types of resources are put into something like `src/main/build/` but am on the fence about that as there is some expectation based on the maven nomenclature and it's opinionated standards that tree is well defined given how it's plugins have expectations (such as filtering is out of `src/main/resources/`, etc. In additionally, there were already existing icon, icns, png images in `src/main/resources/images/` and copying them elsewhere lead to duplication which can lead to error proneness (TM on the word I just invented!)

build.gradle
```diff
@@ -138,86 +143,95 @@ dependencyUpdates {
...
-            '--icon', './native/rails.png',
-            '--file-associations', './native/rails-linux-association.properties',
-            '--app-version', appVersion(),
-            '--input', input,
-            '--dest', output,
-            '--main-jar', shadowJar.archiveName,
-            '--linux-package-name', 'rails-18xx',
-            '--linux-app-category', 'games',
-            '--linux-shortcut'
+    doFirst {
```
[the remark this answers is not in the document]
jklap: Re "testing" ... is there a way you can look at the previous PR CI runs (haven't used GitHub much so not sure). But in any case, running `gradle clean test` is what failed (the exec buiding tasks worked fine). `doFirst` (and `doLast`) aren't used for defining multiple actions in a task, they are used to define when that closure executes. The issue we ran into here is that when you define code within the "body" of the task, it is effectively parsed during the gradle configuration step-- even if the task ends up not being run. So in this case, even though we were executing only `clean` and `test`, the `jpackage()` method was being called during the gradle configuration step and hence building the `jpackage` path and with the addition of the jpackage exists check it started failing for non-JDK 14 runs. Moving the code to `doFirst` means that closure won't execute unless the task is actually executing, hence it won't make the `jpackage()` method call. Not sure this is really the best/most appropriate solution but it works and I didn't really want to spend much more time on it ;) If you'd like to research further and fine the correct approach or validate this approach is "correct" that would be awesome. Or we can just punt this as it seems to work fine until if/when we move to use of a jpackage plugin for building..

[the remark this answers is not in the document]
madoar: I'm fine with this change. I just didn't understand why you added it therefore my question.

About your question:

> Re "testing" ... is there a way you can look at the previous PR CI runs (haven't used GitHub much so not sure).

Behind each commit you have either a red cross or a green check. If you click on one of these you will get to the results of the GitHub Actions workflows for the corresponding commit.

src/main/resources/build/last_rails-osx-association.properties
madoar: I agree that we shouldn't add redundancy, but were we using the icons in the Rails application at all? I'm fine with adding them to the `src` folder, but if we only use them for the installers then `main/resources` does not fit in my opinion. Maybe we should add them to `src/installers` or something similar?

build.gradle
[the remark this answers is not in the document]
jklap: Yep, found it earlier today after I wrote the above.

I did comment yesterday about why I added it but I recollect that gradle is new(ish) to you so my comment may not have been clear enough without a deeper gradle background. Sorry for that!

src/main/resources/build/last_rails-osx-association.properties
jklap: I didn't/haven't looked to see if/where they may be used. I did glance at the files I removed (ie for the old installer approach) and didn't see any references there so I assumed they were in use elsewhere. Didn't bother looking deeper than that.

PR #164, opened 2020-04-15
enable support for double clicking saved game/last rails on Mac

jklap:
Enables support for handling a double click of a saved game (or last rails) file on a Mac.

May also work on Windows and Linux (but don't have those platforms to test)

Renamed the `.last_rails` file to `.lrails` due to Mac not liking the `_` in the file extension (it doesn't seem to recognize the file association). There was an open question in Slack regarding the file extension but no decision so I just went with this to get the PR out

### DIFF
--- a/src/main/java/net/sf/rails/util/RunGame.java
+++ b/src/main/java/net/sf/rails/util/RunGame.java
@@ -1,38 +1,54 @@
 package net.sf.rails.util;
 
+import java.awt.*;
+import java.awt.desktop.OpenFilesEvent;
+import java.awt.desktop.OpenFilesHandler;
 import java.io.File;
 
+import javax.swing.*;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import net.sf.rails.common.ConfigManager;
+import net.sf.rails.ui.swing.AutoSaveLoadDialog;
 import net.sf.rails.ui.swing.GameSetupController;
 
 public class RunGame {
 
-    public static void main(String[] args) {
+    private static final Logger log = LoggerFactory.getLogger(RunGame.class);
 
-        // Initialize configuration
-        ConfigManager.initConfiguration(false);
-        
-        int nargs = 0;
+    public static void main(String[] args) {
+        final String[] fileName = { null };
+
         if (args != null && args.length > 0) {
-            nargs = args.length;
-            System.out.println("Number of args: "+nargs);
             for (String arg : args) {
                 System.out.println ("Arg: "+arg);
             }
+            fileName[0] = args[0];
         }
 
+        if ( Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.APP_OPEN_FILE)) {
+            Desktop.getDesktop().setOpenFileHandler(e -> {
+                for ( File file : e.getFiles() ) {
+                    // open the file
+                    fileName[0] = file.toString();
+                    // break out as we only handle one game at a time...
+                    break;
+                }
+            });
+        }
+
+        // Initialize configuration
+        ConfigManager.initConfiguration(false);
+
         // currently we only start one game
-        if (nargs >= 1) {
-            loadGame(args[0]);
+        if ( fileName[0] != null ) {
+            GameLoader.loadAndStartGame(new File(fileName[0]));
         } else {
             /* Start the rails.game selector, which will do all the rest. */
             GameSetupController.Builder setupBuilder = GameSetupController.builder();
             setupBuilder.start();
         }
-    }
-
-    private static void loadGame (String filePath) {
-        File gameFile = new File(filePath);
-        GameLoader.loadAndStartGame(gameFile);
     }
 }
--- a/src/main/resources/build/last_rails-osx-association.properties
+++ b/src/main/resources/build/last_rails-osx-association.properties
@@ -1,4 +1,4 @@
-extension=last_rails
+extension=lrails
 mime-type=application/rails-18xx-last
 icon=./src/main/resources/images/icon/last_rails.icns
 description=Rails polling file
--- a/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
+++ b/src/main/java/net/sf/rails/ui/swing/GameUIManager.java
@@ -105,7 +105,7 @@ public class GameUIManager implements DialogOwner {
     protected static final String DEFAULT_SAVE_PATTERN = "yyyyMMdd_HHmm";
     public static final String DEFAULT_SAVE_EXTENSION = "rails";
 
-    public static final String DEFAULT_SAVE_POLLING_EXTENSION = "last_rails";
+    public static final String DEFAULT_SAVE_POLLING_EXTENSION = "lrails";
     protected static final String NEXT_PLAYER_SUFFIX = "NEXT_PLAYER";
     protected static final String CURRENT_ROUND_SUFFIX = "CURRENT_ROUND";
 


src/main/java/net/sf/rails/util/RunGame.java
```diff
@@ -1,38 +1,54 @@
...
 import net.sf.rails.common.ConfigManager;
+import net.sf.rails.ui.swing.AutoSaveLoadDialog;
 import net.sf.rails.ui.swing.GameSetupController;
 
 public class RunGame {
 
+    private static final Logger log = LoggerFactory.getLogger(RunGame.class);
+
     public static void main(String[] args) {
+        final String[] fileName = { null };
```
madoar, line 22: Can we replace this with a `List<String>`? I'm not a friend of arrays containing `null` values

src/main/resources/build/last_rails-osx-association.properties
```diff
@@ -1,4 +1,4 @@
-extension=last_rails
+extension=lrails
```
madoar, line 1: I'm in favor of using `.lastrails` :)

src/main/java/net/sf/rails/util/RunGame.java
jklap: This type of usage is common when you need to reference & change/set a variable within an anonymous class where it's a single item, ie String, Integer/int, etc. Creation of a List is confusing as it leads the developer/reader to think there may be multiple items. I'm also of the mind set that creation of a List (ie the object creation) to hold a single item is a code smell and is overhead not needed. In this case, I don't see there is much difference (ir it was possible) between setting a String to null and setting the first element in the array to null. 

Maybe it's just from my background of C/C++ but nulls in arrays aren't a big deal to me :)

src/main/java/net/sf/rails/util/RunGame.java
```diff
@@ -1,38 +1,54 @@
...
+
+        if ( Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.APP_OPEN_FILE)) {
+            Desktop.getDesktop().setOpenFileHandler(e -> {
+                for ( File file : e.getFiles() ) {
+                    // open the file
+                    fileName[0] = file.toString();
+                    // break out as we only handle one game at a time...
+                    break;
+                }
+            });
```
madoar, line 39: Is this executed in a blocking manner? 
If not this will not work, because you can't be sure that this is executed before the parts after the `if` condition

jklap: On the Mac this executes immediately after the handler is installed, hence `fileName[0]` is set before the `if` is hit. I don't know about other platforms. And yes, this does work on the Mac; I've been using it for a couple of weeks now although it works much better now with your jpackage changes as the previous `dmg` creation required me to update the `Info.plist` manually with the file associations after I created the `dmg`.. hence making my build and testing process quite tedious. Finally wrote a shell script stringing together the commands and `patch` and can not ditch that

src/main/java/net/sf/rails/util/RunGame.java
madoar: Ah I understand now. Yes lambda functions can only access effectively `final` variables.

> Maybe it's just from my background of C/C++ but nulls in arrays aren't a big deal to me :)

For me it is quite the opposite. I've done a lot with `scala` and there you try to not use `null` when possible. Instead you can use `Option` with its two flavors `Some` and `None` to symbolize whether a value is set or not.

In Java I think it is fine to make use of `null` as long as it does not looks too crazy :)

src/main/java/net/sf/rails/util/RunGame.java
madoar: I think that the `OpenFileHandler` is executed in the Swing thread and not in your normal application thread. Therefore it should be random that you got your expected result.
In addition an introduction in this topic from [Oracle](https://docs.oracle.com/javase/tutorial/deployment/selfContainedApps/fileassociation.html) suggests that this approach only works for Mac. For Linux and Windows you get the opened file as an input argument.

As a possible alternative it may be worth it to take a look at how this can be achieved with JavaFX, especially if we want to migrate more of our UI to JavaFX.

src/main/java/net/sf/rails/util/RunGame.java
jklap: Of course after this whole discussion I rewrote this logic anyway (for a different reason) in an upcoming PR ;)

Java has something similar, and in fact I think they stole the idea from Scala, `Optional`, although I personally don't see much value in it as even with it you still end up having to make an `if` check, it's just supposedly a bit prettier because you call `isPresent()`. Again, probably my C/C++ background but I think eyeballing logic and seeing that `== null` is clearer to me then my brain having to process the fact that a method is being called, and that method is basically making the `null` check when I see `isPresent()`. Maybe it makes more sense for developers that haven't had extensive experience with `null` in other languages but for me, blah.

src/main/java/net/sf/rails/util/RunGame.java
jklap: Not sure how JavaFX would change our handling of receiving open events from the OS? I don't know much about JavaFX (in fact my only exposure is what we have in Rails) so any pointers would be appreciated for reading.

Either way, until we determine whether we would pursue anything JavaFX as an alternative, this approach does work for the bundled Mac app :)

madoar: > Not sure how JavaFX would change our handling of receiving open events from the OS? I don't know much about JavaFX (in fact my only exposure is what we have in Rails) so any pointers would be appreciated for reading.

I also have no experience with JavaFX + file associations. A quick search led me to [this](https://stackoverflow.com/a/52606228) StackOverflow question. I think it may be a good starting point. 

> Either way, until we determine whether we would pursue anything JavaFX as an alternative

I agree. The part about a JavaFX solution was meant as a long-term goal.

> ..., this approach does work for the bundled Mac app

I still think that the `OpenFileHandler` is called from a separate thread, which makes this solution very unpredictable and unstable.

You could check this by printing the thread inside the `OpenFileHandler` and outside it by accessing `Thread.currentThread()`.

jklap: Read the Stackoverflow link-- that may work for Linux and Windows based on the comments regarding how they are passed the file but will not work on the Mac as opening files are never passed as an argument.

Over the last several weeks of using this I've seen no random behavior-- it has *always* opened the file double clicked on.

madoar: > Read the Stackoverflow link-- that may work for Linux and Windows based on the comments regarding how they are passed the file but will not work on the Mac as opening files are never passed as an argument.

I don't know if this will work on Mac. It is just meant as a starting point. The long-term problem with the `Desktop` solution is that it is `awt` so if we want to move from `swing`/`awt` to JavaFX long-term we will either need to replace this with something that purely JavaFX or we accept a mix of JavaFX and `swing`/`awt`.

But this is not a question we must face now, it is simply meant as a topic which we can think about in the future.

> Over the last several weeks of using this I've seen no random behavior-- it has always opened the file double clicked on.

I can understand that this has always worked for you. But that an issue didn't have a visible impact does not mean that it is not there right? To exclude the possibility that the `OpenFileHandler` is executed from a separate thread I would really appreciate it if you could test whether the thread that calls the `OpenFileHandler` is equal to the one that executes:

https://github.com/Rails-18xx/Rails/blob/27cc1391e455f1e10bdbf3293e78de4e4a1b4614/src/main/java/net/sf/rails/util/RunGame.java#L42-L52

If this is the case then everything as it should be, otherwise we know that we can face a race condition.

jklap: As I believe I mentioned previously-- this code has already been changed to add another feature so am unable to spend time to backtrack and try to test for an edge case that you may believe is there that will be a non-issue once this PR has been merged and I can submit the following PR

jklap: But DO welcome your review on the following PR for similar....

madoar: Ok

PR #165, opened 2020-04-15
support for archiving old game save files

jklap:
This PR adds support for archiving off old saved game files. This adds 3 configuration options:
* Archive enabled?
* Archive directory
* Retain count

It works by building up a list of Rails saved game files in the directory the current saved game is found in, leaving the most recent "count" of them alone and moving the rest into the specified archive directory. The directory can be specified by either a full path (in which case you'd end up with different game's saved game files all mingled together (which is likely not wanted behavior) or if not absolute, then it is relative to the directory the saved game is found in. If this is not specified then the default is `./archive/`.

The primary use case of this would be use of auto-load/save polling where the files continue to build up (quickly) over time as every player takes their turn. Moving these older files out keeps the list of saved games short so you can quickly find the most recent if you need to (re)open it in addition to keeping the old files in case you want/need to rollback.


src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -943,12 +955,74 @@ protected boolean save(GameAction saveAction) {
...
+            } else if ( archiveDirFile.exists() && ! archiveDirFile.isDirectory() ) {
+                log.warn("Archive directory doesn't seem to be a directory?");
+                archive = false;
+            }
+
+            if ( archive ) {
+                // iterate through files in current directory
+                SortedSet<File> files = new TreeSet<>((a, b) -> ComparisonChain.start()
+                        .compare(b.lastModified(), a.lastModified())
+                        .result());
```
madoar, line 1001:
```suggestion
                SortedSet<File> files = new TreeSet<>(Comparator.comparing(File::lastModified));
```

jklap: Your suggestion won't work-- if you'll notice in the original code we are comparing `b` to `a` to generate the sorted list in descending order and then pick up our iteration after `count` items (ie the # of items we leave/retain) thereby avoiding an array "location" check on every pass. Your suggestion would sort them in ascending order.

jklap:
> I'm not sure about mixing `Path` and `File`. Is it possible to use only one of them?

Not sure what the concern here is? Use of the new `Path`/`Paths`/`Files` in conjunction with `File` is completely appropriate. The newer (well, added in JDK 7) `Paths` and related are to support working with file paths and to simplify dealing with file & directory manipulation (iteration, monitoring, etc) and aren't really designed to replace `File` nor is `File` as capable as the new objects in that type of manipulation or is `File` being replaced by the new classes.

src/main/java/net/sf/rails/game/GameManager.java
madoar: So the order needs to be reversed?

Then we could use:

```suggestion
SortedSet<File> files = new TreeSet<>(Comparator.comparing(File::lastModified).reversed());
```

madoar:
In my opinion the `nio` API (the one with `Path` and `Files`) can be seen as being distinct from the older `File` API. Both APIs can be used to achieve the same things without needing to mix them.

My point is that (correct me if I'm wrong) you are using the newer `nio` API only in a single place (when creating the directory). Why not replace this with the older `File` API? If you do this it will be more uniform.

As a note: this is only a suggestion.

src/main/java/net/sf/rails/game/GameManager.java
jklap: Corresponding change should also then be made in `GameSetupController` given this is where this code originated from :)

Either way, I believe from a previous conversation/PR that the general consensus was a preference for `ComparisonChain` vs `Comparator`? So unless there is another compelling reason to make this change I'm going to hold off until this code (and `GameSetupController`) is touched again.

madoar: As far as I remember we had no consensus yet. The only people discussing were us two right?

The question in my opinion is:
For which functionality do we want to use external libraries?

A lot of functionality that previously required using an external library are now contained directly in Java (for example a comfortable way to create `Comparator` objects). This leads me to the question whether we should still use external libraries for such things? 

About my reason to mark these lines:
I'm only marking this here because you're touching this code. 
This is the best moment where we can think about changing something.
For me it is kind of irrelevant if you copy the code, reformat it or change it in some other way as long as you touch it.

jklap:
> In my opinion the `nio` API (the one with `Path` and `Files`) can be seen as being distinct from the older `File` API. Both APIs can be used to achieve the same things without needing to mix them.
> 
> My point is that (correct me if I'm wrong) you are using the newer `nio` API only in a single place (when creating the directory). Why not replace this with the older `File` API? If you do this it will be more uniform.


Well it is a good question-- why the mixing... the primary reason I used this here:
```
Files.createDirectories(Path.of(archiveDir,File.separator,"dummy"));
```

Is that it is more tolerant and has fewer lines of code to handle edge cases such as your file being in a top level directory. For example, to create all parent directories the following code is commonly used:
```
new File("<whatever>").getParent().mkdirs();
```

Yet this code will fail with a NPE if the file is a top-level file as there is no parent. So then you need to introduce a null check on `getParent()` usually via `if`.

I personally prefer the newer classes for most things as the `Path` and `Paths` classes have some really useful methods that you used to have to use Apache Commons stuff to one-line. But of course most everything that works with files in the JDK uses a `File` so mixing is pretty much a guarantee.

src/main/java/net/sf/rails/game/GameManager.java
jklap: I believe @neutronc also weighed in.

> For me it is kind of irrelevant if you copy the code, reformat it or change it in some other way as long as you touch it.

I don't think it is irrelevant though-- if we have multiple ways of doing the same things in the code base, ie picking on this case, ordering a set, then it becomes harder for 1) someone to be able to understand the code (because they keep seeing multiple ways of doing the exact same thing) and 2) know what is the best/preferred/suggested way when they are going to write or change code.

In this specific case, we already have lots of places that are using `ComparisonChain` AND `Comparator` so until we have a decision on which we should be using as a standard/best practice, I don't see any value in picking over over the other. If this was completely new code that would be one thing and hence more debatable but as this was existing logic I brought it over as is.

But you also bring up a good question on third party libraries and frameworks. I am personally a fan of using JDK provided functionality only if it's fully equivalent to what we can get from some of the third party libraries. In most cases, for example, a lot of the Apache Commons libraries (text, lang3, commons-io) provide functionality a heck of a lot more feature rich then the closest JDK alternatives-- just look the string substitution for a good comparison. Often these types of decisions (a vs b) tend to be more of a religious nature type discussion & decision based on an individuals own experience and exposure vs one actually better than the other and so to move forward I think we go with what we have unless there is a compelling reason to switch -- and at which point it should be a task to go though the code based and make that switch-- not to add a new library/whatever and just use for new code but to actually migrate everything.

madoar: I would like to hear @neutronc opinion about this, so that we find a final answer to this question.

madoar:
> But of course most everything that works with files in the JDK uses a File so mixing is pretty much a guarantee.

Which APIs of the JDK do you mean? I can't think of an API that requires you to use a `File` object over a `Path` object. I agree that this applies to most third party libraries, but this is most likely the case because of how old they are.

src/main/java/net/sf/rails/game/GameManager.java
jklap: Why don't you raise this up on Slack with both examples and pursue the answer there as I believe @blentz mentioned previously not everyone watches here. I don't really care one way or the other.

neutronc: Tough nut, but from this example :  I like the comparator  though i would need to get used to the notation. what i dislike in this example that the explicit Treeset is lacking information about its objects thats confusing for me as non development trained guy :)
So i would vote to go with comparator.

PR #167, opened 2020-04-15
remove RailsRoot static instance

jklap:
This PR builds on #120 

The primary focus of this PR is to replace `RailsRoot` as a singleton/static instance. In most places the change was pretty clean but there were some `Action` classes that made references to the singleton and hence needed the `RailsRoot` passed in during object creation. In addition there were some changes to the deserialization logic but those already had access to the `RailsRoot` instance via the stream but instead were referencing it via the static method.

This PR also includes some other static variable/instance cleanups that were more minor in nature and seemed to be historical in nature and not really needing to be static with the current code base.

Note: this PR does *not* change all `Action` classes to include `RailsRoot` passed in during construction, only those that needed it. This could (should?) be a future PR such that we standardize those `Action` classes.

FYI I have 1 or 2 more PRs coming to "finish" cleanup of singleton/static usages were appropriate (ie still leaving things that are appropriate singletons such as the `ConfigurationManager`). Those PRs are currently "blocked" from submission by some of the other currently open PRs I have.

### DIFF
--- a/src/main/java/net/sf/rails/algorithms/NetworkGraphModifier.java
+++ b/src/main/java/net/sf/rails/algorithms/NetworkGraphModifier.java
@@ -1,29 +1,32 @@
 package net.sf.rails.algorithms;
 
 import net.sf.rails.game.PublicCompany;
+import net.sf.rails.game.RailsRoot;
 
 /**
  * Classes that change properties of the network graph
  * before both route evaluation and revenenue calculation starts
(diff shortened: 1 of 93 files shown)


neutronc:
I do refrain from reviewing this. @blentz can you please share your thought about that ?

jklap:
Thanks for the review @madoar, this one has lots to poke through

PR #173, opened 2020-04-18
static report window & buttons learned about myTurn

jklap:
This PR adds a configuration option that makes the report window "static" in that you can not click on a link to rollback/undo to that point.

In addition, it fixes a bug when loading a saved game in that the top of the report window displays the timewarp active buttons and text even though you aren't actually in a timewarp.

Finally the report window learns about myTurn and disallows undo when it is not your turn (but should still allow undo via moderator). This addresses an issue where someone executes an undo after their turn has finished and the file was written at which point the other users may have loaded the file and hence causes load issues when the previous player redoes their turn.

If a user really needs to rollback their turn they should coordinate it with the other players and re-open their original game file and start from there.

### DIFF
--- a/src/main/java/net/sf/rails/common/ReportBuffer.java
+++ b/src/main/java/net/sf/rails/common/ReportBuffer.java
@@ -1,19 +1,20 @@
 package net.sf.rails.common;
 
+import java.util.Deque;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
+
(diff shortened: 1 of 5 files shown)


jklap:
Because it becomes problematic when playing with auto-load/saving as a mis-click on a URL can cause a game file that has been saved to be changed and hence disrupting the loading on the other players.

Don't disagree that it is a great feature but when playing in real-time it causes more issues then benefits as the ability to use the normal undo/redo buttons within a user's turn can accomplish everything that the window provides yet without the downsides.

It also looks like there used to be a static window previous due to the existing configuration options and some commented out code but may have been removed at some point due to refactoring so I just (re)added the support :)

PR #174, opened 2020-04-20
general cleanup of deserialization methods

jklap:
Some general cleanup of the deserialization methods. This PR contains:
* some minor single line (re)formatting
* some TODO notes in areas that needed further review for auto-load/save handling (detailed notes forthcoming in a new PR of the specific issue post merge of this PR)
* cleanup of RailsRoot references that aren't needed due to parent

### DIFF
--- a/src/main/java/rails/game/action/ExchangeTokens.java
+++ b/src/main/java/rails/game/action/ExchangeTokens.java
@@ -82,11 +82,4 @@ public class ExchangeTokens extends PossibleORAction {
         ;
     }
 
-    /** Deserialize */
-    private void readObject(ObjectInputStream in) throws IOException,
-            ClassNotFoundException {
-
-        in.defaultReadObject();
-
-    }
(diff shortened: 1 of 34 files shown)


src/main/java/rails/game/action/ExchangeTokens.java
```diff
@@ -82,11 +82,4 @@ public String toString() {
...
     }
 
-    /** Deserialize */
-    private void readObject(ObjectInputStream in) throws IOException,
-            ClassNotFoundException {
-
-        in.defaultReadObject();
-
-    }
 }
```
neutronc, line 85: Hm no replacement isnt hat Object needed ?

jklap: `readObject()` is only needed if you are going to customize the reading-- in this case we are just turning around and calling the default reader.

PR #178, opened 2020-04-20
auto-load/reload serialization fixes

jklap:
This PR addresses several issues (really the same type of issue) that can occur while use of auto-load/save. The issue is specifically due to how we construct an action internally which is then stored in the execution history vs how it is deserialized and then (reloaded and) compared against that action history. In several cases, even when replaying (and hence comparing) one's own actions, changes are found due to differences in (usually) "empty" objects such as an empty `List` or empty `Map` whereas the action, for game play purposes, is really the same. This mostly has to do with how and when we are constructing arrays, lists and maps but in theory could happen with other object types too. This PR only addresses cases there it is clear when & how an object is created (and hence its variables) vs when it is deserialized. It is likely there are additional scenarios out there not addressed by the PR.

Before this PR we were running into auto-load failures frequently (between every 10-20 turn changes); post this change we have not yet seen a single auto-load failure (several hundred turn changes). 

I will add my detailed notes I made while troubleshooting this in a comment below, keep in mind I was "talking to myself" as I made the notes and so may be only make sense to me... but also adding it here for future reference for myself :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,6 +72,8 @@ dependencies {
     implementation 'org.apache.httpcomponents:httpclient:4.5.12'
     implementation 'org.apache.commons:commons-lang3:3.10'
     implementation 'org.apache.commons:commons-text:1.8'
+    // https://www.jetbrains.com/help/idea/annotating-source-code.html
+    implementation 'org.jetbrains:annotations:16.0.2'
 
     testImplementation 'junit:junit:4.13'
     testImplementation 'org.easytesting:fest-assert-core:2.0M10'
--- a/src/main/java/net/sf/rails/game/specific/_1835/PrussianFormationRound.java
+++ b/src/main/java/net/sf/rails/game/specific/_1835/PrussianFormationRound.java
@@ -73,8 +73,7 @@ public class PrussianFormationRound extends StockRound {
         mergePr = !prussianIsComplete(gameManager);
 
         ReportBuffer.add(this, LocalText.getText("StartFormationRound", PR_ID));
-        log.debug("StartPr="+startPr+" forcedStart="+forcedStart
-                +" mergePr="+mergePr+" forcedMerge="+forcedMerge);
+        log.debug("StartPr={} forcedStart={} mergePr={} forcedMerge={}", startPr, forcedStart, mergePr, forcedMerge);
 
         step = startPr ? Step.START : Step.MERGE;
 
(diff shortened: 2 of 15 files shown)


jklap:
Possible action created with null `locations`
But was loaded and ended up empty `ArrayList`

Deserialize is setting to `new ArrayList()`
Curious: so why don’t we run into this more often?

Comparison between actions is:
```
boolean options = (Objects.equal(this.locations, action.locations) || this.locations == null && action.locations.isEmpty()) && Objects.equal(this.specialProperty, action.specialProperty);
```

`this.locations` is an empty array
`action.locations` is null
So this would fail-- BUT it seems this is a side effect of the issue and the comparison is not the real problem; while we could address it here, need to find where this mismatch is created

Also: shouldn't this compare actually also support flipping of the null and empty comparison?
Also 2: might not actually care if the root cause is addressed

log data:
```
04-09-2020:23:10:18.557 [AWT-EventQueue-0] DEBUG n.s.r.g.GameManager->process():775 - Jason may: Jason may LayBaseToken, company = KO, locations = NULL, specialProperty = NULL, type = 1
```

```
04-09-2020:23:10:21.776 [AWT-EventQueue-0] DEBUG n.s.r.u.s.GameUIManager->processOnServer():420 - ==Passing to server: Jason executed LayBaseToken, company = KO, locations = NULL, specialProperty = NULL, chosenHex = MapHex{uri=/Map/G4}Coordinates{4, 7}, type = 1, chosenStation = 
```

```
04-09-2020:23:19:22.791 [AWT-EventQueue-0] WARN  n.s.r.g.GameManager->reload():1010 - loaded action Jason executed LayBaseToken, company = KO, locations = [], specialProperty = NULL, chosenHex = MapHex{uri=/Map/G4}Coordinates{4, 7}, type = 1, chosenStation = 1 is not the same as expected game action Jason executed LayBaseToken, company = KO, locations = NULL, specialProperty = NULL, chosenHex = MapHex{uri=/Map/G4}Coordinates{4, 7}, type = 1, chosenStation = 1
```

Bug only affects auto-load/reload when the current user actions on `LayBaseToken` and then we attempt to reload (and hence) compare previously executed actions. In this case the `LayBaseToken` action is generated in memory with a `NULL` `locations` property but when it is deserialized from disk it is created with an empty `List`. This comparison then fails. This also shows an issue in the comparison method in that it checks `src.locations == null || target.locations ` is empty but in this case it’s actually `src.locations` is empty and `target.locations == null`. Quick scan of the code shows that the deseralization issue may occur in other places too given when/how arrays are created. This issue does not affect loading a file fresh as there is no existing in-memory action comparisons happening.

In reviewing other classes with similar readObject() logic, the solution is one of:
* Make sure the variable is initialized in the declaration (not constructor as they are not called)
* Make sure the variable is initialized in the readObject() method (non-conditionally)

The choice of which one of this is more appropriate is based on the usage of the class and it’s callers. For instance with ReachDestination is makes sense to initialize in the declaration as it simplifies other logic (ie checking for null and initializing, reduces caller logic for null checks) and is not passed or initialized via it’s constructor. This is vs MergeCompanies which requires non-null instances to be passed in via the constructor which, if we initialized the variable in declaration means we are allocating an object that is immediately replaced. In this case it makes sense to make sure to initialize the variable in readObject()

PR #181, opened 2020-04-21
addition of portfolio manager

jklap:
Additional of a `PortfolioManager` to address some `TODO` comments moving the portfolio lookups out of the `GameManager` and into `RailsRoot`

### DIFF
--- a/src/main/java/net/sf/rails/game/PlayerManager.java
+++ b/src/main/java/net/sf/rails/game/PlayerManager.java
@@ -1,6 +1,8 @@
 package net.sf.rails.game;
 
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/rails/game/PlayerManager.java
```diff
@@ -100,6 +101,11 @@ public void setPlayers(List<String> playerNames, Bank bank) {
 
         ReportBuffer.add(this, LocalText.getText("PlayerCash", cashText));
         ReportBuffer.add(this, LocalText.getText("BankHas", Bank.format(this, bank.getCash())));
+
+        // sets initial priority player and certificate limits
+        priorityPlayer.set(playerModel.playerOrder.get(0));
+        int startCertificates = playerCertificateLimits.get(playerModel.playerOrder.size());
+        playerCertificateLimit.set(startCertificates);
```
neutronc, line 108: Be careful with the positioning, sometimes information are not available before the finishConfiguration Call !

jklap: I looked. This is what the logic looked like initially:
```
         playerManager.initPlayers(gameData.getPlayers(), bank);
         gameManager.init();
         // TODO: Can this be merged above?
         playerManager.init();
```

We are simply merging the logic that was in `playerManager.init()` into `playerManager.initPlayers()`. The only implication would be if there was logic inside of `gameManager.init()` that could affect this and I didn't see anything when I looked. Note too that the above code is actually being called inside of `finishConfiguration()` :)

Issue #130, opened 2020-04-08
certificate exchange in 1835: how can we help?


bearerer:
> 1835 bugs
After president change there is no way to select the certificates to exchange (this is done by the previous director), partial fix is implemented: prefer 10% to 5%

This bug is currently breaking our games. Will there be a fix in the near future? Can we help? (Two developers with a bit of time at hand since Corona keeps us at home anyway).

neutronc:
any help is appreciated if you know your way around in Java :)
But to reconstruct the bug i need a save game

neutronc:
i can give you an introduction in the code base via discord/twitch if you are interested, just lets organise a time and day.
OTherwise fork the repo here at Github, The code you are looking for is in src.main.java.net.sf.rails.game.Stockround.java and
src.main.java.net.sf.rails.game.specific._1835.Stockround_1835.java
Functions

kepakiano:
Hey, I'll chime in on this. I'm the second dev @bearerer mentioned. I just created this savegame using the current rails_2_develop branch. Peter has 2x10% BA, Hans has 2x10% and the President's share. Jan has 2x10% and 1x 20%. It's Hans' turn. If he dumps the presidency by selling 10% BA, the presidency goes to Jan. Jan has 2x10% and 1x20%. Hans should be able to choose which one of the two packages he gets to make up for his 20% ("Der alte Direktor bestimmt die Stückelung"). However, he is given 2x10%.
[Hans_sells_BA.zip](https://github.com/Rails-18xx/Rails/files/4460858/Hans_sells_BA.zip)

neutronc:
Just give me a sutiable time slot (i am on CEST) and in home office work :)

neutronc:
And yes i know about the Problem with the replacement shares, that simply hasnt been implemented.

We can discuss this during the code Intro if you guys would like.

neutronc:
i'll add you guys to the team and grant you access to the repository.

kepakiano:
Would tomorrow work for you? I'd be available all day.

neutronc:
Ok lets schedule a meeting then tomorrow 5pm CEST ?
And btw we have a slack channel, i'll post the invite link here again.

neutronc:
https://join.slack.com/t/rails-18xx/shared_invite/zt-dcq83imv-uFnJ8h5LqyOLmZ66eRFTXw

bearerer:
Sounds good. How do we meet up?

[...]

PR #184, opened 2020-04-23
Fix Issue#130 by adding specific implementation of executePresidentTransferAfterDump for 1835

kepakiano:
https://github.com/Rails-18xx/Rails/issues/130

We're not 100% sure about the line
`Portfolio.moveAll(sellableCertificateCombinations.last(), bankTo);`
We have tested it with all kinds of companies (1x 20, 2x20, 3x20 and the Prussian) in 1835 though and all worked as expected. 

Also, we did not incorporate this solution with the code in `PlayerShareUtils` since we do not know the specific needs of 1880, which uses this function as well. Maybe you can give us some insight on how we can share more code between the games for this issue. We made a method for this, since there is no need for the share swap to include the bank at all in 1835.

### DIFF
--- a/src/main/java/net/sf/rails/game/specific/_1835/StockRound_1835.java
+++ b/src/main/java/net/sf/rails/game/specific/_1835/StockRound_1835.java
@@ -7,20 +7,17 @@ package net.sf.rails.game.specific._1835;
 import java.util.List;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
-import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Sets;
-import com.google.common.collect.SortedMultiset;
-
+import net.sf.rails.game.financial.*;
+import net.sf.rails.game.model.CertificatesModel;
 import rails.game.action.BuyCertificate;
 import rails.game.action.NullAction;
 import net.sf.rails.common.LocalText;
 import net.sf.rails.common.ReportBuffer;
 import net.sf.rails.game.*;
-import net.sf.rails.game.financial.PublicCertificate;
-import net.sf.rails.game.financial.StockRound;
-import net.sf.rails.game.financial.StockSpace;
-import net.sf.rails.game.financial.PlayerShareUtils;
 import net.sf.rails.game.model.PortfolioModel;
 import net.sf.rails.game.state.Portfolio;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/rails/game/specific/_1835/StockRound_1835.java
```diff
@@ -278,5 +275,43 @@ protected void setPriority(String string) {
...
+        }
+
+        // Transfer the sold certificates
+        Portfolio.moveAll(certsToSell, bankTo);
+    }
+
+    private void executePresidentTransferAfterDump(PublicCompany company, Set<PublicCertificate> certsToSell, Player newPresident, int presSharesToSell, Player oldPresident, BankPortfolio bankTo) {
+        PublicCertificate presidentCert = company.getPresidentsShare();
+
+        SortedSet<PublicCertificate.Combination> newPresidentsReplacementForPresidentShare = CertificatesModel.certificateCombinations(
```
jklap, line 301: `Combination` is unnecessarily qualified with `PublicCertificate`

madoar:
It seems like `real/1835_Stoll_12` failed:

```
Added TestGame test/1851_Late
Could not create test fail report
[ERROR] Tests run: 26, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 8.037 s <<< FAILURE! - in net.sf.rails.test.TestGameBuilder
[ERROR] real/1835_Stoll_12(net.sf.rails.test.TestGame)  Time elapsed: 0.357 s  <<< ERROR!
java.lang.NullPointerException
	at net.sf.rails.test.TestGame.gameReportTest(TestGame.java:58)
	at net.sf.rails.test.TestGame.runTest(TestGame.java:39)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   TestGame.runTest:39->gameReportTest:58 NullPointer
[INFO] 
[ERROR] Tests run: 167, Failures: 0, Errors: 1, Skipped: 0
```

kepakiano:
Okay, found out why the file fails. Sorry for not checking earlier.
It crashes, because Marco wants to buy shares from the pool which are not there, since we changed what happens when the presidency changes hands. It plays differently from the point where Eyal sells OL shares. Before that he has: 20% (D), 20% and 10% = 50%. Marco has 2x 10% and 20% = 40%. The game says he sells two 10% certificates, which isn't possible, but maybe that's just a logging issue. He ends up selling a 20% certificate. After he does so, the new solution does the following: Swap his president's share with Marco's 20% share and move the 20% share into the bank. That's how I interpret the rules. Technically, he could have chosen between Marco's 20% certificate and Marco's 2 10% certificate, but when implementing we think getting one cert instead of two is almost always the better option.
The old implementation however would move Marco's two 10% certs into the bank and then move the president's share to Marco. So Eyal would move certs into the bank he never had any control over.

What confused me while implementing this was that the transfer of the sold share has not taken place yet when the function gets called. The only information we get is how many shares get sold to dump the presidency (presSharesToSell) and which certificates need to get sold after this (certsToSell). I would like to just get certsToSell with all the certificates that get sold, move them to the bank and handle the presidency swapping after. Is that possible somehow?

src/main/java/net/sf/rails/game/specific/_1835/StockRound_1835.java
kepakiano: That's true, I'll directly import `Combination`. Thanks!

neutronc:
Has been used and fixed with PR#204, Thanks !

PR #188, opened 2020-04-29
SonarQube cleanup (batch 1)

jklap:
Fix issues identified by running SonarQube against the code base. This is the first batch of several.

### DIFF
--- a/src/main/java/net/sf/rails/game/GameManager.java
+++ b/src/main/java/net/sf/rails/game/GameManager.java
@@ -775,7 +775,7 @@ public class GameManager extends RailsManager implements Configurable, Owner {
 
         // logging of game actions activated
         for (PossibleAction pa : possibleActions.getList()) {
-            log.debug("{}", pa.toString());
+            log.debug("{}", pa);
         }
 
         return result;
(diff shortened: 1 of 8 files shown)


jklap:
> Are you sure that the null checks can be removed ?

As a general statement, yes :) Is there a specific chunk of code you were looking at/concerned about?

Some of the null checks can be removed because the variables they referred to are declared final hence can't be null. Some null checks were removed because they were already being dereferenced earlier in the same block of code and hence would have NPE'ed earlier. These checks I was a little more hesitant to remove in case the earlier code was later checked BUT in reality this specific concern (ie code later being changed) is really an issue with _any_ code change at all and hence thought it was better to simplify and cleanup the code making it easier to read and understand when something like a `null` check wasn't actually needed. There were likely a couple of other general situations but those were the two off the top of my head I can recollect-- which FYI, this is batch 1 of 4 sets of SonarQube changes and the 4th batch that I just finished has more `null` check cleanups although some of those were a bit tricker and hence required more review.

Note too these changes (and the other batches) are all changes that SonarQube determined via it's static code analysis, it wasn't me just seeing a null check that I thought was ok to remove and removing it :)

PR #189, opened 2020-04-29
Load game replacing current game

jklap:
This PR adds support to load a new game (not reload/autoload) but a completely new game replacing the currently open game.

On a Mac you can't (easily for an end user) open multiple copies of the Rails app if you are playing multiple games in real-time so this adds support so you can at least easily switch between them without having to quit and restart the app as currently is required. Obviously is also useful on other platforms too

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/StatusWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/StatusWindow.java
@@ -10,11 +10,13 @@ import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
 import javax.swing.*;
+import javax.swing.filechooser.FileFilter;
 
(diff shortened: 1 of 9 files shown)


src/main/java/net/sf/rails/ui/swing/StatusWindow.java
```diff
@@ -137,7 +144,21 @@ public void initMenu() {
...
+        fileMenu.add(actionMenuItem);
+
+        actionMenuItem = new ActionMenuItem(LOAD_CMD);
+        actionMenuItem.setActionCommand(LOAD_CMD);
+        actionMenuItem.addActionListener(this);
+        actionMenuItem.setEnabled(true);
+        actionMenuItem.setPossibleAction(new GameAction(gameUIManager.getRoot(), GameAction.Mode.LOAD));
+        fileMenu.add(actionMenuItem);
+
+        actionMenuItem = new ActionMenuItem(LocalText.getText("SAVE"));
```
neutronc, line 161: I would like to have a shortkey for those actions as well :)

jklap: Noted-- have a menu change PR coming, will try to find appropriate keys

PR #190, opened 2020-05-03
no network window config & close windows

jklap:
Added a configuration option to _not_ display the route (ie network graph) window when you click on the company name to calculate possible revenue.

Also added tracking of the route windows that are opened so we can close them when we load a new game

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/ORPanel.java
+++ b/src/main/java/net/sf/rails/ui/swing/ORPanel.java
@@ -130,6 +130,8 @@ implements ActionListener, KeyListener, RevenueListener {
     private RevenueAdapter revenueAdapter = null;
     private Thread revenueThread = null;
 
+    private List<JFrame> openWindows = new ArrayList<>();
+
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/rails/ui/swing/ORPanel.java
```diff
@@ -1472,4 +1482,10 @@ public void setTreasuryBonusRevenue(int orCompIndex2, int bonusAmount) {
 
     }
 
+    public void dispose() {
+        for ( JFrame frame : openWindows ) {
+            frame.dispose();;
```
madoar, line 1487: There is a redundant `;` in this line.

jklap: Will nuke it!

jklap:
@madoar I personally don't use it at all... I find it much to difficult to mentally parse when I'm in the middle of a game. But I do think there are other things that could be done to make it much improved--  including, maybe, even creating it as some type of temp overlay on the map itself instead of a pop up window. But either way, until that day, here is to hiding it!

PR #199, opened 2020-05-05
SonarQube (batch 2)

jklap:
Fixes related to to running code analysis with SonarQube against Rails.

This is the second set of changes-- another set is being prepped for a PR in a day or two.

These three sets of changes address many of the bug/critical/high issues identified by SonarQube but not all (I can only handle doing these in batches ;)

### DIFF
--- a/src/main/java/rails/game/correct/CorrectionManager.java
+++ b/src/main/java/rails/game/correct/CorrectionManager.java
@@ -93,6 +93,9 @@ public abstract class CorrectionManager extends RailsAbstractItem {
 
 
     public boolean equals(Object object) {
+        if ( ! (object instanceof CorrectionManager) ) {
+            return false;
+        }
         CorrectionManager cm = (CorrectionManager) object;
         return this.getParent() == cm.getParent() && this.correctionType == cm.correctionType;
     }
--- a/src/main/java/rails/game/correct/CorrectionManager.java
+++ b/src/main/java/rails/game/correct/CorrectionManager.java
@@ -92,7 +92,8 @@ public abstract class CorrectionManager extends RailsAbstractItem {
     }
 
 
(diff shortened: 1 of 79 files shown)


src/main/java/rails/game/correct/CorrectionManager.java
```diff
@@ -92,7 +92,8 @@ protected boolean execute(CorrectionAction action) {
     }
 
 
-    public boolean equals(CorrectionManager cm) {
+    public boolean equals(Object object) {
+        CorrectionManager cm = (CorrectionManager) object;
```
madoar, line 96: This is a dangerous assumption. 
You should add a type check here

jklap: Yes, yes it is. I was working off the assumption that any `equals()` in place were already invalid due to the parameter declaration of `CorrectionManager` but that would have actually caused the default `equals()` to be called vs the override. So yeah, this should be fixed.

jklap: LOL, this is what is in my current code base:
```
    public boolean equals(Object object) {
        if ( ! (object instanceof CorrectionManager) ) {
            return false;
        }
...
```

Evidently I already fixed it!

neutronc: As Madoar said please fix :)

jklap: I've already got this in the following commit-- I'll have to rebase and edit that commit to pull this out :(

I looked through the code and don't believe this `equals()` method is actually being executed and looking at code coverage from unit tests it is certainly not. Given that I was just going to sit on it for the next PR given the pain of rebasing.

PR #201, opened 2020-05-06
SonarQube

jklap:
Batch 3.

Changes focused around:
* Removal of `toString()` in logger calls
* Change from `Map.keySet() -> Map.get(..)` to use of `Map.entrySet()`
* Change from `stringVar.equals("blah")` to `"blah".equals(stringVar)` to eliminate NPE concerns
* Change keyword order to be consistent (and match recommended) (ie like `abstract protected` to `protected abstract`)
* Remove unnecessary boxing and unboxing (ie `Integer.valueOf()` or `new Integer()`)

### DIFF
--- a/src/main/java/net/sf/rails/algorithms/NetworkTrain.java
+++ b/src/main/java/net/sf/rails/algorithms/NetworkTrain.java
@@ -34,7 +34,7 @@ public final class NetworkTrain implements Comparable<NetworkTrain>{
         this.isETrain = isETrain;
         this.trainName = trainName;
         this.railsTrain = train;
-        log.debug("Created NetworkTrain {} / {}", this.toString(), this.attributes());
+        log.debug("Created NetworkTrain {} / {}", this, this.attributes());
     }
 
     static NetworkTrain createFromRailsTrain(Train railsTrain){
@@ -63,10 +63,10 @@ public final class NetworkTrain implements Comparable<NetworkTrain>{
         boolean ignoreTowns = false; int multiplyCities = 1; int multiplyTowns = 1;
         boolean isHTrain = false;
         boolean isETrain = false;
-        if (t.equals("D")) {
+        if ( "D".equals(t)) {
             log.debug("RA: found Diesel train");
             cities = 99;
-        } else if (t.equals("TGV")) {
+        } else if ( "TGV".equals(t)) {
             log.debug("RA: found TGV  train");
             cities = 3;
             ignoreTowns = true;
--- a/src/main/java/net/sf/rails/algorithms/NetworkVertex.java
+++ b/src/main/java/net/sf/rails/algorithms/NetworkVertex.java
@@ -260,12 +260,12 @@ public final class NetworkVertex implements Comparable<NetworkVertex> {
         // define locationName
         stopName = null;
         if (station.getType() == Station.Type.OFFMAPCITY) {
-            if (hex.getStopName() != null && !hex.getStopName().equals("")) {
+            if (hex.getStopName() != null && !"".equals(hex.getStopName())) {
                 stopName = hex.getStopName();
             }
         } else {
-            if (hex.getStopName() != null && !hex.getStopName().equals("")
-                    && station.getStopName() != null && !station.getStopName().equals("")) {
+            if (hex.getStopName() != null && !"".equals(hex.getStopName())
+                    && station.getStopName() != null && !"".equals(station.getStopName())) {
                 stopName = hex.getStopName() + "." + station.getStopName();
             }
         }
@@ -307,7 +307,7 @@ public final class NetworkVertex implements Comparable<NetworkVertex> {
     public static final class ValueOrder implements Comparator<NetworkVertex> {
 
         public int compare(NetworkVertex vA, NetworkVertex vB) {
-            int result = -((Integer)vA.getValue()).compareTo(vB.getValue()); // compare by value, descending
+            int result = -Integer.compare(vA.getValue(), vB.getValue()); // compare by value, descending
             if (result == 0)
                 result = vA.compareTo(vB); // otherwise use natural ordering
             return result;
--- a/src/main/java/net/sf/rails/algorithms/NetworkVertex.java
+++ b/src/main/java/net/sf/rails/algorithms/NetworkVertex.java
@@ -7,6 +7,7 @@ import java.util.*;
 import net.sf.rails.game.*;
 import net.sf.rails.ui.swing.hexmap.*;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.jgrapht.graph.SimpleGraph;
@@ -260,12 +261,11 @@ public final class NetworkVertex implements Comparable<NetworkVertex> {
         // define locationName
         stopName = null;
         if (station.getType() == Station.Type.OFFMAPCITY) {
-            if (hex.getStopName() != null && !"".equals(hex.getStopName())) {
+            if ( StringUtils.isNotBlank(hex.getStopName()) ) {
                 stopName = hex.getStopName();
             }
         } else {
-            if (hex.getStopName() != null && !"".equals(hex.getStopName())
-                    && station.getStopName() != null && !"".equals(station.getStopName())) {
+            if ( StringUtils.isNotBlank(hex.getStopName()) && StringUtils.isNotBlank(station.getStopName())) {
                 stopName = hex.getStopName() + "." + station.getStopName();
             }
         }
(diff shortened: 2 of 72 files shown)


src/main/java/net/sf/rails/algorithms/NetworkVertex.java
```diff
@@ -260,12 +260,12 @@ public boolean initRailsVertex(PublicCompany company) {
         // define locationName
         stopName = null;
         if (station.getType() == Station.Type.OFFMAPCITY) {
-            if (hex.getStopName() != null && !hex.getStopName().equals("")) {
+            if (hex.getStopName() != null && !"".equals(hex.getStopName())) {
```
madoar, line 263: Maybe better do a size comparison instead? Or use a method from apache-commons

src/main/java/net/sf/rails/algorithms/NetworkVertex.java
```diff
@@ -260,12 +260,12 @@ public boolean initRailsVertex(PublicCompany company) {
...
                 stopName = hex.getStopName();
             }
         } else {
-            if (hex.getStopName() != null && !hex.getStopName().equals("")
-                    && station.getStopName() != null && !station.getStopName().equals("")) {
+            if (hex.getStopName() != null && !"".equals(hex.getStopName())
```
madoar, line 267: Maybe better do a size comparison instead? Or use a method from apache-commons

src/main/java/net/sf/rails/algorithms/RevenueAdapter.java
```diff
@@ -375,14 +375,13 @@ private int maxTravelEdges() {
...
         }
-        for ( NetworkEdge edge : edgeTravelSets.keySet() ) {
-            EdgeTravel edgeTravel = edgeTravelSets.get(edge);
+        for ( Map.Entry<NetworkEdge, EdgeTravel> entry : edgeTravelSets.entrySet() ) {
             StringBuilder edgeString = new StringBuilder("RA: EdgeSet for ").
-                    append(edge.toFullInfoString()).
+                    append(entry.getKey().toFullInfoString()).
                     append(" size = ").
-                    append(edgeTravel.set.size()).
+                    append(entry.getValue().set.size()).
```
madoar, line 382: I agree using `Set#entrySet` is better here. Still I think we should give both the key and the value a better name before working with them by adding something similar to:

```java
final NetworkEdge edge = entry.getKey();
final EdgeTravel edgeTravel = entry.getValue();
```

src/main/java/net/sf/rails/game/CompanyManager.java
```diff
@@ -269,16 +269,14 @@ public Company getCompany(String type, String name) {
...
         for (PrivateCompany priv : lPrivateCompanies) {
             if (priv.isCloseable()) // check if private is closeable
                 priv.setClosed();
         }
     }
 
     public List<PrivateCompany> getPrivatesOwnedByPlayers() {
-        List<PrivateCompany> privatesOwnedByPlayers =
-                new ArrayList<PrivateCompany>();
+        List<PrivateCompany> privatesOwnedByPlayers = new ArrayList<PrivateCompany>();
```
madoar, line 279:
```suggestion
        List<PrivateCompany> privatesOwnedByPlayers = new ArrayList<>();
```

src/main/java/net/sf/rails/game/CompanyType.java
```diff
@@ -70,12 +70,11 @@ public void finishConfiguration (GameManager gameManager) {
...
     throws ConfigurationException {
-        Company newCompany = null;
+        Company newCompany;
         try {
             newCompany = Configure.create(Company.class, className, this, id);
         } catch (Exception e) {
-            throw new ConfigurationException(LocalText.getText(
-                    "ClassCannotBeInstantiated", className), e);
+            throw new ConfigurationException(LocalText.getText("ClassCannotBeInstantiated", className), e);
         }
```
madoar, line 78: Note for later:

Maybe we should change this to something similar to:

```java
try {
   final Company newCompany = Configure.create(Company.class, className, this, id);

   // do something with newCompany
} catch (Exception e) {
   throw new ConfigurationException(LocalText.getText("ClassCannotBeInstantiated", className), e);
}
```

src/main/java/net/sf/rails/game/StartItem.java
```diff
@@ -336,8 +336,8 @@ public int getBid(Player player) {
      */
     public int getBidders() {
         int bidders = 0;
-        for (Player bidder : active.keySet()) {
-            if (active.get(bidder).value() == true) {
+        for ( Map.Entry<Player, BooleanState> entry  : active.entrySet()) {
+            if ( entry.getValue().value() ) {
                 bidders++;
             }
         }
```
madoar, line 343:
```suggestion
        for (BooleanState booleanState : active.values()) {
            if (booleanState.value()) {
                bidders++;
            }
        }
```

src/main/java/net/sf/rails/game/StartItem.java
```diff
@@ -390,8 +390,8 @@ public boolean isActive(Player player) {
      * participate in an auction (e.g. 1862)
      */
     public void setAllActive() {
-        for (Player p : active.keySet()) {
-            active.get(p).set(true);
+        for ( Map.Entry<Player, BooleanState> entry : active.entrySet()) {
+            entry.getValue().set(true);
         }
```
madoar, line 395:
```suggestion
        for (BooleanState booleanState : active.values()) {
            booleanState.set(true);
        }
```

src/main/java/net/sf/rails/ui/swing/elements/Spinner.java
```diff
@@ -11,8 +11,7 @@
     private Color buttonColour = new Color(255, 220, 150);
 
     public Spinner(int value, int from, int to, int step) {
-        super(new SpinnerNumberModel(new Integer(value), new Integer(from),
-                to > 0 ? new Integer(to) : null, new Integer(step)));
+        super(new SpinnerNumberModel(Integer.valueOf(value), Integer.valueOf(from), to > 0 ? to : null, Integer.valueOf(step)));
```
madoar, line 14: Autoboxing?

src/main/java/net/sf/rails/algorithms/NetworkTrain.java
```diff
@@ -63,10 +63,10 @@ public static NetworkTrain createFromString(String trainString) {
         boolean ignoreTowns = false; int multiplyCities = 1; int multiplyTowns = 1;
         boolean isHTrain = false;
         boolean isETrain = false;
-        if (t.equals("D")) {
+        if ( "D".equals(t)) {
```
madoar, line 66: A note for a future changes:
These strings look like constants to me that should be places in `public final static` Strings

jklap: A lot of these might even be better represented by a `switch`

src/main/java/net/sf/rails/ui/swing/elements/Spinner.java
jklap: Yep. I'm sure there are lots of changes like this that still need to be made!

src/main/java/net/sf/rails/algorithms/NetworkVertex.java
jklap: Yeah, would be better to move this to using something like `StringUtils.isNotBlank()`. I didn't really spend time to review the code for a better way to implement, just migrated it

src/main/java/net/sf/rails/algorithms/RevenueAdapter.java
jklap: Man, I had to introduce a variable if it's only used once (as in here). I do see the value if it's used multiple times... do you really think we should still change this?

src/main/java/net/sf/rails/ui/swing/elements/Spinner.java
jklap: Actually I had looked at this one-- can't change it because the autoboxing to `Integer` triggers the `SpinnerNumberModel(Number value, Comparable<?> minimum, Comparable<?> maximum, Number stepSize)` constructor vs the `SpinnerNumberModel(int value, int minimum, int maximum, int stepSize)` constructor.

src/main/java/net/sf/rails/game/specific/_1837/StockRound_1837.java
```diff
@@ -93,6 +93,7 @@ protected void setGameSpecificActions() {
                 for (PublicCompany majorComp : comps) {
                     if ((comp.isRelatedToNational(majorComp.getId()))
                         && (majorComp.hasStarted())
+                            // FIXME: getType() will never equal "Major"
```
neutronc, line 96: the context information is wrong as we are looking for all PublicCompanies which may be a major or a minor... The check is valid.

jklap: `majorComp.getType().equals("Major")`

The problem with this conditional check is that `getType()` returns a `CompanyType` object-- hence `CompanyType.equals("Major")` is the method that is actually being called. Since a `String` is being passed into the `equals()` it will never be true

PR #202, opened 2020-05-08
fixed issue with reload StartCompany action for 18EU

jklap:
This PR addresses an issue with auto-loading via polling (or reload) in that the selected home station is not set in the deserialized object (it's lazily set) so when we go to compare actions during the auto-load it fails. Loading the game fresh does not show this issue as we do not do executed action vs loaded action comparisons.

### DIFF
--- a/src/main/java/rails/game/specific/_18EU/StartCompany_18EU.java
+++ b/src/main/java/rails/game/specific/_18EU/StartCompany_18EU.java
@@ -5,6 +5,11 @@ import java.io.ObjectInputStream;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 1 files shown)


src/main/java/rails/game/specific/_18EU/StartCompany_18EU.java
```diff
@@ -180,7 +189,8 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
                 availableHomeStations.add(hex.getRelatedStop(stationId));
             }
         }
-        // selectedHomeStation is delayed
```
neutronc, line 183: did you check in a real game that this doesnt break the mechanismn ?

jklap: What do you mean in a real game? I've been playing with this change for several days now-- since this only affects deserialization I would think that the loading and auto-loading I've been doing to play would validate?

PR #206, opened 2020-05-10
Base Tests: Recreated Test Report files necessary due to Changes in StockRound and ShareSellingRound

neutronc:


madoar:
Can you please elaborate a bit more what you changed inside the tests to make them work again? This would help me to better understand what your change does :)

neutronc:
just rebuild them
the tests are nothing else then save games that are automatically played by the engine and the report is compared line by line with the prepared report

PR #208, opened 2020-05-12
Use PlayerManager in StartRoundWindow

madoar:
This PR replaces the `players` array and `np` field with the globally available `PlayerManager` model.
This changes is a first step to a single place where the "game state" is defined.


madoar:
@neutronc @jklap I have only tested this with 1835 could you please give this PR a go with other variants to see if it works for these as intended too?

PR #215, opened 2020-05-14
Use PlayerManager in GridPanel and its subclasses

madoar:
This PR introduces `PlayerManager` to `GridPanel` and its subclasses.
It represents the second part of #208


madoar:
@neutronc the class `GameStatus` has a field called `actorIndex`:

https://github.com/Rails-18xx/Rails/blob/a65a7c8bc01eb616999cf42e4a50dcc707f34ab1/src/main/java/net/sf/rails/ui/swing/GameStatus.java#L141-L144

This field looks to me like it could be replaced with a `players.getCurrentPlayer()`, but I'm not entirely sure because of the mentioning of "Company (from treasury)" in the comment above. Can you shed some light here?

madoar:
@jklap @neutronc again I have only tested this PR with 1835. It would be nice if you could give it a test with some other variants so that we are sure it works as intended :)

PR #216, opened 2020-05-14
Use StartRound in StartRoundWindow

madoar:
This PR directly uses `StartRound` to fetch the `StartItem`s instead of querying them only once during initialization.

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
@@ -49,6 +49,16 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
     private static final Color soldColour = new Color(128, 128, 128);
     private static final Color defaultColour = Color.BLACK;
 
+    protected static final String[] itemStatusTextKeys =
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
```diff
@@ -178,18 +177,15 @@ public void init(StartRound round, GameUIManager parent) {
 
         players = gameUIManager.getRoot().getPlayerManager();
 
-        packet = round.getStartPacket();
-        crossIndex = new int[packet.getNumberOfItems()];
+        crossIndex = new int[round.getNumberOfStartItems()];
```
madoar, line 180: In my tests with 1835 the changed line leads to the same result as the initial code.
But I'm not sure whether this is always the case. @neutronc do you know if both methods, i.e. `round.getStartPacket().getNumberOfItems()` should lead to the same result as `round.getNumberOfStartItems()`?

madoar:
Similar to #208 and #216 I have only tested this PR with 1835 therefore it would be useful if for example @neutronc and @jklap could this this PR with other variants of 18xx as well :)

PR #217, opened 2020-05-16
pr-mbr/abstracting-SpecialTileLay

neutronc:
Introducing an abstract SpecialTileLay and its primary Implementation the SpecialSingleTileLay
with the necessary changs of the CompanyManager.xml data

### DIFF
--- a/src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
+++ b/src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
@@ -1,0 +1,114 @@
+package net.sf.rails.game.special;
+
+import net.sf.rails.common.LocalText;
+import net.sf.rails.common.parser.ConfigurationException;
+import net.sf.rails.common.parser.Tag;
+import net.sf.rails.game.*;
+import net.sf.rails.util.Util;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class SpecialSingleTileLay extends SpecialTileLay {
+    /**
+     * Used by Configure (via reflection) only
+     *
+     * @param parent Railsitem this belongs
+     * @param id this Item has
+     */
+    public SpecialSingleTileLay(RailsItem parent, String id) {
+        super(parent, id);
+    }
+
+    @Override
+    public void configureFromXML(Tag tag) throws ConfigurationException {
+        super.configureFromXML(tag);
+
+        Tag tileLayTag = tag.getChild("SpecialSingleTileLay");
+        if (tileLayTag == null) {
+            throw new ConfigurationException("<SpecialSingleTileLay> tag missing");
+        }
+
+        locationCodes = tileLayTag.getAttributeAsString("location");
+        if (!Util.hasValue(locationCodes))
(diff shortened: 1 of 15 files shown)


neutronc:
A Second class with additional features for MultiTileLay handlings will be added in another PR.
Feedback and comments appreciated.

src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
```diff
@@ -0,0 +1,164 @@
...
+
+        if (tileId != null) {
+            description = LocalText.getText("LayNamedTileInfo",
+                    tileId,
+                    name != null ? name : "",
+                            locationCodes,
+                            (extra ? LocalText.getText("extra"):LocalText.getText("notExtra")),
+                            (free ? LocalText.getText("noCost") : discount != 0 ? LocalText.getText("discount", discount) :
+                                LocalText.getText("normalCost")),
+                            (connected ? LocalText.getText("connected") : LocalText.getText("unconnected"))
```
madoar, line 59: Can we split this in multiple statements? This is quite a long statement I think it would be easier to understand if it is split in multiple statements

src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
```diff
@@ -0,0 +1,164 @@
...
+            );
+        } else {
+            description = LocalText.getText("LayTileInfo",
+                    locationCodes,
+                    (tileColours != null ? Arrays.toString(tileColours).replaceAll("[\\[\\]]", ""): ""),
+                    (extra ? LocalText.getText("extra"):LocalText.getText("notExtra")),
+                    (free ? LocalText.getText("noCost") : discount != 0 ? LocalText.getText("discount", discount) :
+                        LocalText.getText("normalCost")),
+                    (connected ? LocalText.getText("connected") : LocalText.getText("unconnected"))
+            );
```
madoar, line 69: See above

src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    public boolean isExecutionable() {
+        return true;
+    }
+
+    @Override
+    public boolean isExtra() {
+        return extra;
+    }
```
madoar, line 105: If the fields are specified in the parent class it is better to also implement these getter methods in the parent class

neutronc: I do agree

src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
neutronc: Thats just the original implementation move down in the singleclass, but yes...

src/main/java/net/sf/rails/game/special/SpecialSingleTileLay.java
jklap: This would be my only comment too

PR #218, opened 2020-05-17
Cleanup StartRoundWindow

madoar:
This PR cleans up the class `StartRoundWindow` in preparation to replace it with a JavaFX implementation.

In addition it seems like the usage of `crossIndex = new int[round.getNumberOfStartItems()];` was incorrect. I fixed this by changing the line to `crossIndex = new int[round.getStartPacket().getNumberOfItems()];`

### DIFF
--- a/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
@@ -98,8 +98,6 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
 
     private PlayerManager players;
 
-    private StartItemAction[] actionableItems;
-
     private int[] crossIndex;
     protected StartRound round;
     private GameUIManager gameUIManager;
@@ -184,8 +182,6 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
 
             crossIndex[item.getIndex()] = i;
         }
-
-        actionableItems = new StartItemAction[round.getNumberOfStartItems()];
 
         infoIcon = createInfoIcon();
 
@@ -409,8 +405,8 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
 
         for (i = 0; i < round.getNumberOfStartItems(); i++) {
             setItemNameButton(i, false);
-            actionableItems[i] = null;
         }
+
         // Unselect the selected private
         dummyButton.setSelected(true);
 
@@ -456,7 +452,7 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
         for (StartItemAction action : actions) {
             j = action.getItemIndex();
             i = crossIndex[j];
-            actionableItems[i] = action;
+
             item = action.getStartItem();
 
             if (action instanceof BuyStartItem) {
--- a/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
+++ b/src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
@@ -94,7 +94,7 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
     private SpinnerNumberModel spinnerModel;
     private ActionButton passButton;
 
-    private ImageIcon infoIcon = null;
+    private ImageIcon infoIcon;
 
     private PlayerManager players;
 
@@ -103,15 +103,12 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
     private GameUIManager gameUIManager;
 
     // For the non-modal dialog to ask for a company starting share price.
-    protected JDialog currentDialog = null;
-    protected PossibleAction currentDialogAction = null;
-    protected SortedSet<StockSpace> startSpaces = null;
-
-    private StartItem si;
-    private JComponent f;
+    protected JDialog currentDialog;
+    protected PossibleAction currentDialogAction;
+    protected SortedSet<StockSpace> startSpaces;
 
     protected PossibleActions possibleActions;
-    protected PossibleAction immediateAction = null;
+    protected PossibleAction immediateAction;
 
     private final ButtonGroup itemGroup = new ButtonGroup();
     private ClickField dummyButton; // To be selected if none else is.
@@ -155,8 +152,7 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
             bidButton.setEnabled(false);
             buttonPanel.add(bidButton);
 
-            spinnerModel =
-                    new SpinnerNumberModel(999, 0, null, 1);
+            spinnerModel = new SpinnerNumberModel(999, 0, null, 1);
             bidAmount = new JSpinner(spinnerModel);
             bidAmount.setPreferredSize(new Dimension(50, 28));
             bidAmount.setEnabled(false);
@@ -175,7 +171,7 @@ public class StartRoundWindow extends JFrame implements ActionListener, KeyListe
 
         players = gameUIManager.getRoot().getPlayerManager();
 
-        crossIndex = new int[round.getNumberOfStartItems()];
+        crossIndex = new int[round.getStartPacket().getNumberOfItems()];
 
         for (int i = 0; i < round.getNumberOfStartItems(); i++) {
             final StartItem item = round.getStartItem(i);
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
```diff
@@ -293,66 +285,58 @@ private void initCells() {
...
-                    si);
-            addField(f, infoXOffset, infoYOffset + i, 1, 1, WIDE_LEFT);
+            info[i] = new Field(infoIcon);
+            info[i].setToolTipText(getStartItemDescription(si));
+            HexHighlightMouseListener.addMouseListener(info[i], gameUIManager.getORUIManager(), si);
+            addField(info[i], infoXOffset, infoYOffset + i, 1, 1, WIDE_LEFT);
 
             // Invisible field, only used to hold current item status.
-            f = itemStatus[i] = new Field(si.getStatusModel());
+            itemStatus[i] = new Field(si.getStatusModel());
```
madoar, line 327: What is meant here by "invisible"? As I see it this field is never added to the UI, therefore I propose removing it.

This field is currently used to store a value which is later parsed into an int. I think we should replace this usage with a direct call to `si.getStatusModel()`

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
```diff
@@ -361,56 +345,45 @@ private void initCells() {
...
 
         for (int i = 0; i < np; i++) {
-            f = lowerPlayerCaption[i] = new Field(players.getPlayerByPosition(i).getPlayerNameModel());
-            addField(f, playerFreeCashXOffset + i, playerFreeCashYOffset + 1,
-                    1, 1, WIDE_TOP);
+            lowerPlayerCaption[i] = new Field(players.getPlayerByPosition(i).getPlayerNameModel());
+            addField(lowerPlayerCaption[i], playerFreeCashXOffset + i, playerFreeCashYOffset + 1, 1, 1, WIDE_TOP);
         }
 
         dummyButton = new ClickField("", "", "", this, itemGroup);
```
madoar, line 359: Similar to the field `itemStatus` this field does not seem to be added to the UI instead it is only added to a button group and preselected at the beginning of `updateStatus`. What is this button used for, do we require it?

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
jklap: I think the `Field` name is misleading here-- while yes, it is a `JLabel`, it's actually implementing `Observer` -- key to note as it means whenever the underlying `IntegerState` is changed the value stored in the `JLabel` is changed, which is then retrieved in `setItemNameButton()`.

So creation of something like an int array and storing off the result won't work-- it won't be updated.

So either we store off the `IntegerState` returned from the `si.getStatusModel()` call or we create our own observer to watch over that `IntegerState`. 

I'm not sure keeping the reference to `si` is appropriate as it that is actually a change in logic-- we assume `si` is valid long term but it very well may not be.

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
jklap: It is being added to the UI-- it's added via the `addField()` method on line 356 and it's manipulated during game play (highlighting) like this:

Screen Shot 2020-05-18 at 6 13 06 PM

jklap:
Regarding you `crossIndex` comment in the description of the PR-- what is the difference between those two methods you mention?

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
neutronc: Interesting Picture there is already missing a field ? whats the item 1?

neutronc: Furthermore by removing the actions you are removing the replayability i think.

jklap: @neutronc that is 18EU... the minors are numbered, not named. The picture is correct ;)

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
madoar: In the end this will most likely be replaced by an `IntegerBinding` in the UI class. If possible I would refrain from using bindings or observers/observables in the model/state classes long-term because they tend to be highly linked to UI implementations. 

Ideally I would like to change our UI to induce its own (UI) updates in an explicit way, but we are not there yet.

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
madoar: You misunderstood my comment.

I didn't mean `lowerPlayerCaption`, I meant the `ClickField` object `dummyButton`.
According to my IntelliJ this field is only accessed twice in the Code:
- in line 359 where it is defined
- in line 388 where it is set as selected

Except these two occurrences the field is never used.

madoar:
> Furthermore by removing the actions you are removing the replayability i think.

@neutronc which actions do you mean?
If you mean the `actionableItems` array then it is only used in thrice in the code according to my IntelliJ:
- in line 188 to define the array with a given size
- in line 412 to set one of its inner values to `null`
- in line 459 to set one of its inner values to `action`

Similar to `dummyButton` this field is otherwise never accessed, especially in a read operation.

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
jklap: Seems it's purpose is clarified by the comment on line 387 and comment on line 114?

Now whether it's actually need to accomplish what those two comments indicate is the question-- which you'd have to manually test I think to verify.

jklap:
`dummyButton` is actually used :) 

But for the topic of `actionableItems`...

I don't see the purpose of `actionableItems`. It's a `private` so can't be used by subclasses as it's not exposed via any getter-- it's only created and elements set-- never read. 

BUT it may be worth retaining as I could see it useful for a variant's subclass of `StartRoundWindow` as it would give that class access to the underlying `StartItemAction`s used.

madoar:
@jklap `actionableItems` should be equal to `possibleActions.getType(StartItemAction.class)`, where `possibleActions` is set in the `init(...)` method. Therefore I think it should be safe to replace its read access occurrences with `possibleActions.getType(StartItemAction.class)`.

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
madoar: Yes I guess I don't get around testing this.

For now I think it is safe to ignore the `dummyButton` field and let it remain part of the class.
I think I will remove the field when we change the window to a JavaFX implementation.

madoar:
> Regarding you crossIndex comment in the description of the PR-- what is the difference between those two methods you mention?

This is a good question, which I'm not too sure about. As far as I undestand it:
- `round.getNumberOfStartItems` fetches you the total number of start items that your game contains. I think this may be independent of whether they are already sold or not. I think these start items don't contain not yet available start items, but I may be wrong with this assumption
- `round.getStartPacket().getNumberOfItems` returns the number of items in the start packet. According to the JavaDoc for `getStartPacket` the start packet contains all items that the current player may bid upon. I think this list may be a subset of all start items, which the first method queries

src/main/java/net/sf/rails/ui/swing/StartRoundWindow.java
neutronc: Pleas remember that you also need to change and test all inherited Classes with Game Specific implementations.
And dont assume that there are no changes in certain fields that are static in the base game (like the player action order in the startroundWindow) Subsequent StartRounds might have a different PlayerOrder (1880 Investor is an example).

madoar: > Pleas remember that you also need to change and test all inherited Classes with Game Specific implementations.

I need your help testing all the game specific implementations, because I don't know how they should behave. I need people that know how these game modes should work to evaluate whether this PR contains any regressions in this regard. Therefore it is also fine if this PR takes a while until it is merged.

> And dont assume that there are no changes in certain fields that are static in the base game (like the player action order in the startroundWindow) Subsequent StartRounds might have a different PlayerOrder (1880 Investor is an example).

I normally don't assume stuff. I base my changes on observations in the code. If I'm not sure whether a change is correct I ask other developers about their opinion (for example see https://github.com/Rails-18xx/Rails/pull/216#discussion_r425347239). 

If you see additional changes you think are dangerous in anyway and should be further analyzed please comment on them. This comment is about an unchanged line which because it is unchanged does not lead to a regression.

PR #225, opened 2020-05-19
1844 : Current State of Implementation of 1844 from the archive rails…

neutronc:
…2-code

Introduced flippable trains from Erik Vos
Introduced HoldingCompany for the private MountainRailRoad and TunnelCompany (seels at a fix price dont operate but are worth something if activated before at game end.


neutronc:
Superseeded by #226

PR #226, opened 2020-05-19
1844 : Current State of Implementation of 1844 from the archive rails…

neutronc:
Current Implementation of 1844 rebased on the new rails_2_development branch

Introduced flippable trains from Erik Vos
Introduced HoldingCompany for the private MountainRailRoad and TunnelCompany (sells at a fix price dont operate but are worth something if activated before at game end).


src/main/java/net/sf/rails/game/financial/ShareSellingRound.java
```diff
@@ -25,7 +25,7 @@
 
 // TODO: Check if un-initialized states cause undo problems
 public class ShareSellingRound extends StockRound {
-    private static final Logger log = LoggerFactory.getLogger(ShareSellingRound.class);
+    protected static final Logger log = LoggerFactory.getLogger(ShareSellingRound.class);
```
jklap, line 28: Should be private

src/main/java/net/sf/rails/game/Phase.java
```diff
@@ -93,6 +93,10 @@
     private ImmutableList<TrainCertificateType> releasedTrains;
     private String releasedTrainNames;
 
+    /** Dual train types to flip (turned upside down) if a phase gets activated */
+    protected List<TrainType> flippedTrains;
+    String flippedTrainNames;
```
jklap, line 98: Can this be `private`?

src/main/java/net/sf/rails/game/PublicCompany.java
```diff
@@ -35,7 +35,7 @@
  */
 public class PublicCompany extends RailsAbstractItem implements Company, RailsMoneyOwner, PortfolioOwner, Comparable<PublicCompany> {
 
-    private static final Logger log = LoggerFactory.getLogger(PublicCompany.class);
+    protected static final Logger log = LoggerFactory.getLogger(PublicCompany.class);
```
jklap, line 38: `private`

src/main/java/net/sf/rails/game/TrainCertificateType.java
```diff
@@ -206,6 +206,11 @@ public String getInitialPortfolio() {
         return initialPortfolio;
     }
 
+    public boolean isDual() {
+        if (potentialTrainTypes.size() == 2) return true;
```
jklap, line 210: This could be written as thereby avoiding the if/else
```
return potentionalTrainTypes.size() == 2;
```

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
```diff
@@ -527,12 +528,61 @@ public void updateStatus(boolean myTurn) {
             gameStatus.setPriorityPlayer(gameUIManager.getPriorityPlayer().getIndex());
         }
 
+        // Must Special menu be enabled?
+        boolean enabled = updateSpecialActionMenu();
+        specialMenu.setOpaque(enabled);
+        specialMenu.setEnabled(enabled);
+        specialMenu.repaint();
+
+        passButton.setEnabled(false);
```
jklap, line 537: Should these really be here? Do you remember why you added this here as we have logic triggered via `GameUIManager` to handle the refresh/state of pass/autoPass

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
```diff
@@ -527,12 +528,61 @@ public void updateStatus(boolean myTurn) {
...
+                    default:
+                        break;
+                }
+            }
+        }
+
+        if (currentRound instanceof EndOfGameRound) endOfGame();
+        pack();
+
+        toFront();
```
jklap, line 573: Note to self-- this needs to be migrated to `GameUIManager`s `toFront()`

src/main/java/net/sf/rails/ui/swing/StatusWindow.java
jklap: Actually the buttons were already disabled above at line 485 so I'm assuming this is a leftover from merging?

neutronc: Old Code most likely, need to check again

src/main/java/net/sf/rails/game/PublicCompany.java
neutronc: Hm why ? theres a PublicCompany_1844 needing access to that logger...

src/main/java/net/sf/rails/game/financial/ShareSellingRound.java
neutronc: Cant be, inherited by a class that needs access to that logger...

src/main/java/net/sf/rails/game/Phase.java
neutronc: yes

src/main/java/net/sf/rails/game/TrainCertificateType.java
neutronc: done

src/main/java/net/sf/rails/game/PublicCompany.java
jklap: Loggers should always be private and *only* available to the class declared in. If a subclass needs a logger, it should get it's own :)

The most obvious reason is that if you use a logger from a different class the detail included in the log will be incorrect-- it, for instance, would show the wrong "logger name" giving the indication the log line was generated a class different than it actually was, ie a log line like this would end up misleading:
```
05-18-2020:17:21:32.804 [AWT-EventQueue-0] DEBUG n.s.r.u.s.GameUIManager->processOnServer():612 -  - ==Result from server: true
```

As the creation of the logger is a one-time hit during class construction (not even instance construction) there is no real reason to not declare a logger for each class that needs to log.

PR #231, opened 2020-05-20
18Chesapeake, 1844: First release of needed MultiTileLay SpecialClass

neutronc:
Creating new class SpecialMultiTileLay as 
handling class for Attributes from newer Privates 
like in 1844 or !8Chesapeake just to mention a few.

Comments welcome.

### DIFF
--- a/src/main/java/net/sf/rails/game/special/SpecialMultiTileLay.java
+++ b/src/main/java/net/sf/rails/game/special/SpecialMultiTileLay.java
@@ -1,0 +1,236 @@
+package net.sf.rails.game.special;
+
+import net.sf.rails.common.LocalText;
+import net.sf.rails.common.parser.ConfigurationException;
+import net.sf.rails.common.parser.Tag;
+import net.sf.rails.game.*;
+import net.sf.rails.game.state.IntegerState;
+import net.sf.rails.util.Util;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author Martin Brumm <dr.martin.brumm@t-online.de>
+ * @date May 20th 2020
+ * This Class is based on the assumption that while you can lay mutiple Tiles, each specific tile can only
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/rails/game/special/SpecialMultiTileLay.java
```diff
@@ -0,0 +1,236 @@
...
+ * @author Martin Brumm <dr.martin.brumm@t-online.de>
+ * @date May 20th 2020
+ * This Class is based on the assumption that while you can lay mutiple Tiles, each specific tile can only
+ * be placed once on the map with this SpecialPower for the private with that special power.
+ */
+public class SpecialMultiTileLay extends SpecialTileLay {
+	
+	private String locationCodes = null;
+	private List<MapHex> locations = null;
+	private String name = null;
```
jklap, line 24: `name`, `locations`, `locationCodes`, `extra`, `free`, `discount` and `connected` and `tileColours` all exist in the base class? Assuming we don't really need to declare these here? We can then also ditch the related getters and setters

src/main/java/net/sf/rails/game/special/SpecialMultiTileLay.java
```diff
@@ -0,0 +1,236 @@
...
+    @Override
+    public void configureFromXML(Tag tag) throws ConfigurationException {
+        super.configureFromXML(tag);
+
+        Tag tileLayTag = tag.getChild("SpecialMultiTileLay");
+        if (tileLayTag == null) {
+            throw new ConfigurationException("<SpecialMultiTileLay> tag missing");
+        }
+
+        locationCodes = tileLayTag.getAttributeAsString("location");
```
jklap, line 63: Might be worth moving the "common" configuration such as `name`, `tileColours`, etc out of here and into the parent class `SpecialTileLay` as that same code is also used by the single tile lay class too, correct? Then we would just leave the unique configuration here.

madoar:
How is this PR related to #226?

Issue #234, opened 2020-05-28
18Scan issue: runTo="tokenonly" for an off-map city is not obeyed.


erik-vos:
In 18Scan, off-map cities may only be run to if a token of the running company is present there. This is configured by the runTo="tokenonly" attribute in Map.xml. The attached picture suggests that this runTo value (understandably) has not yet been implemented.
![18Scan_IgnoresTokenOnly](https://user-images.githubusercontent.com/13908049/83169985-20cdd680-a114-11ea-925b-4fe72ac76a16.jpg)

As I am not familiar at all with the routing code, I wonder if anyone can do something about it. I will push the current 18Scan code to erik-vos/18Scan. Here follows a save file.

[18Scan_IgnoresTokenOnly.zip](https://github.com/Rails-18xx/Rails/files/4697115/18Scan_IgnoresTokenOnly.zip)

erik-vos:
I have found the cause: the Map.xml defaults are ineffective because they are parsed _after_ being called during MapHex parsing, so these are all set to null. It's unclear to me how this could have happened. Anyway, so far it does not seem to be a route finding problem, that code looks OK.

I am working on it.

erik-vos:
After a lot of debugging and diagram scribbling I think I have reached the bottom of this issue. At configuration time there are more use-before-defined cases than the one I mentioned above. At execution time it works, if I specify tokenonly per MapHex rather than as an offmap default. It even works too well: even token laying is prevented! So I will have to fix that first.

The background of this all is a discussion I had with Stefan during 2011/12, about which I have resurfaced a few emails. At that time I had implemented a new class Access to deal with these Stop properties. Access has been part of later versions of Rails 1.x; all I have kept of those days is a v1.7.12 source and a v1.9.1 jar, both containing Access. In Rails 2.0, Stefan had renamed Access to StopType, a name that confused me at first, because we are not dealing with types but with properties that (in theory) can be set per individual Tile or MapHex. He also had changed the configuration code pretty heavily, with the above results. 

I'm not going to reintroduce the original Access class, but surely the Stop property config code will get closer to my earlier version of it.

erik-vos:
I have fixed the token laying prevention issue mentioned in my previous post, by adding a new boolean parameter called "running" through 6 layers of network code down to Stop.isRunToAllowed(). It is always true (to show the valid train routes at all times), but false in the ORUIManager code to find valid hexes for tile and token laying.

[...]

Issue #232, opened 2020-05-21
Finalize dual trains


erik-vos:
Emergency buying a dual train certificate does not work, as I discovered in 18Scan. This appears to be due to the way such trains are intialised, therefore some rework is to be done on that matter. I have split it off to a separate branch called Dual Trains.

Just a question for now:
When buying a dual (or "flippable") train card from the Pool or from another company, may that card then be flipped, i.e. buying the train on the other side than the original one? 
My rule books of 1844 and 18Scan seem to be silent on this matter. Has this ever been clarified?

neutronc:
As far as i know the player has the right to choose which train he wants to buy on the dual train certificate. After he has decided the type the Train is finalized and the certificate cant be flipped anymore. This is at least the mechanismn on 1844. Checking 18Scan rules.

neutronc:
18Scan is not specifiying that as well, but so the certificate is representing both train types until the player decides which type he wants and pays the respective money.

erik-vos:
The question is: after a train card has been bought for the first time and the buyer has decided on its type, are future owners of that same card bound to that choice (i.e. when bought from the Pool or from another company). I found that 18VA rules are explicit about this: in the Pool, and only there, becomes the train type undefined again. 

As I want to prepare for that situation as well, the solution will rather deeply affect the whole train handling. One first decision to take is, whether portfolios own _trains_ or _train certificates_. I am undecided, but somewhat leaning to the latter now. On the other hand, you can always find the certificate of a certain train, so perhaps we can keep using the current usage anyhow.

neutronc:
havent thought about that (18VA) but unless its explicitely stated the certificate becomes the train type that the player chooses upon buy from the bank for the first time. I dont want Players to be able to switch certifiate types upon crossselling them to other companies...
Thats a NO GO. buying a 1+1 and selling that as 2 Train to someone else for 100 K making 20k profit on a train sale..

erik-vos:
OK, that's reasonable for the cross-selling case: no flipping.
And how about buying from the Pool? It may be a rare event, but we have to deal with it. From there trains must be always bought at face value. 
It would be nice to know if any other games have rules for that. I just found that in 1812 such trains do _not_ become undecided in the Pool, so that's a a different case then.

My conclusion is, that this behaviour must become configurable rather than hard-coded.

erik-vos:
I have built in a configurable game parameter that specifies if dual trains become undecided in the Pool. Default = false.
Values:
[...]

PR #238, opened 2020-06-08
Dual trains and a partial fix for runTo=tokenonly.

erik-vos:
This PR covers the following:
1. #232 Make dual trains work. New class TrainCard, changes to most train-related code.
2. #234 RunTo="tokenonly" access to offmap cities does not work. Partial fix, issue stays open.
3. Some more initial work on 18Scan, just scratched the surface.
3. Various minor corrections.
All tests are passed.


erik-vos:
> 
> 
> You did remove the NL Tile here ? By Chance or accident ?

I'm not aware of any tiles removed, should be an accident then. Which tile number?

erik-vos:
OK, reintroduced in TileDesigner.18t and the overall Tiles.xml, and pushed to my repo. Will that do?

blentz:
> Edit: it's a separate commit. How to add it to the PR?

As long as the commit is in the branch this PR is based on, it'll be added to the PR automagically.

Issue #255, opened 2020-06-20
Wrong price when force selling 10% Prussian share


madoar:
When `Bjoern` buys a new `6+` train in the attached game save he is missing 224 money and therefore needs to sell shares to pay the full price. In theory it should be enough to sell a `10%` P share for 212 money and a `10%` WT share for 64 money. After selling the `10%` P share the client only subtracts 106 money from the remaining amount instead of the full 212.

Game save: [1835_20200604_1940_Bjoern.rails.zip](https://github.com/Rails-18xx/Rails/files/4807617/1835_20200604_1940_Bjoern.rails.zip)

neutronc:
Good morning, which version ?

madoar:
It occurs in all 2.2.x versions as far as I know.

PR #256, opened 2020-06-20
2.2.1-Maintenance Fix for Issue #255 

neutronc:
Fixing a Bug in ShareSellingRound stemming from the Fact that the original getSellingPrice does not account for shareUnits while StockRound_1835 has a sellingPrice that is checking the shareUnits. This needs to be followed up..


erik-vos:
The PR says that 4 of the 6 checks were unsuccessful. Not sure what that means, perhaps you should look at that first, @neutronc

neutronc:
Test data cant be imported apparently via a pull request :(

PR #274, opened 2020-07-27
New profile properties to re-enable old OR panel displays, and stuff on 18Scan

erik-vos:
This PR adds two new user profile properties which add two new Configuration options under the Appearance tab.

The first is to reactivate the revenue spinner, as announced before.
The second comes from me being highly annoyed by the disappearance of the not-operating companies in the ORPanel except in the train buying step. This new "show all companies" property re-enables just that.

The defaults remain as is, so to activate these options, use Configuration/Appearance in the startup window.

To my surprise the bulk of this PR is stuff from my work on 18Scan; I thought that had already been merged, but apparently not.


erik-vos:
Thanks Martin. This looks like good stuff.

Wat we need is (a) generic method(s) to detect connectivity, in or near the GameManager class. We do not only need this for checking Destinations (1856, 18Scan) in operating rounds, but also in 18EU and 1861 to check if merging minors are connected in stock rounds or special merge rounds, and no doubt there is much more. So this should be generally accessible code.

I'll have to study this code and the Network classes better, but it looks like a very good start. At first, we should take inventory of the requirements. 

But that all is a different subject than this PR.

PR #277, opened 2020-08-09
18Scan completed. 

erik-vos:
Status set to Fully Playable.

This PR adds correct calculation of bonus amounts.
In addition there are some minor corrections.

### DIFF
--- a/src/main/java/net/sf/rails/game/Bonus.java
+++ b/src/main/java/net/sf/rails/game/Bonus.java
@@ -4,10 +4,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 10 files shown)


erik-vos:
It seems that these commits have not made it into upstream/master. When I compare that with my master, these commits are missing. Has anything gone wrong here? 
See also issue #290

neutronc:
the commits are in Rails-18xx:master

neutronc:
did you pull the new master from the origin ?

erik-vos:
No, I pull updates from upstream and push to origin.
Strange. When I compared my local master with upstream/master, the report was that these three commits are missing in upstream/master.
But now I have pulled all new stuff from upstream, and the differences have disappeared. Don't understand, but I rest my case.

PR #294, opened 2020-09-25
Steam over Holland (incomplete)

erik-vos:
Implementing SOH takes longer than I expected. To break the tension I'm preparing this PR, so volunteers can have a look.

See the Game Notes for what is not yet ready, and how to select one of the three different initial round procedures.

Some notes on changes to core functionality:

1. Added "river" hex sides, similar to "impassable", but not blocking tile laying. Same colour, but I'll probably change that. Impact on tile laying cost will be added later.

2. A simple random generator has been added to GameManager and is used for the random private selection of SOH. To make the random generator seed persistent through save/load, a new "hidden" type of GameOption was added. Hidden means that it does not appear as an Option in the game selection window.

3. Starting a company by buying more than one certificate in one turn was implemented by presetting StartCompany.numberBought. The UI will no longer update this value if it is already nonzero.

4. Added moveRight and moveLeft methods to StockMarket for linear markets.


neutronc (changes requested):
Theres are two  temporary statements in MapManager ? Is that still necessary ?

neutronc (commented):
Would it be possible to store the seed and give a player an option to create a seed ? So people can share the seeds and generate the same random setup ?

neutronc (approved):

neutronc (approved):

erik-vos:
Temporary: I think only the comment itself is a leftover and can be removed. Found only one in MapManager, but I can't exclude that there are more leftovers of 18Scan debugging.

Seed: I supposed that the first player starts the game and save it, whereafter the other players load it, picking up the seed that way. But I don't know what happens in practice. What is the "use case"?
If we would allow entering a seed created off-game, we can remove "hidden" and make it a numeric entry field.
We could also report a game-created seed (date/time in milliseconds) in the Report window, so it can be distributed. 
Not sure what you want.

PR #313, opened 2020-12-01
Creating a Ledge for the Stockmarket of 1870 maybe SoH.

neutronc:


neutronc:
grafik

erik-vos:
Checking SOH, it turns out that the ledge above 100 hides the start price value red box marking around that same value.
Without ledges: 
WithoutLedges
With ledges: 
WithLedges

erik-vos:
Thinking this through, wouldn't it be easier, and perhaps look better (also for 1870), if I would replace LeftOfLedge by RightOfLedge (in the next box, of course)?

PR #316, opened 2020-12-02
1880-Fixes-for-2.2.2-Tile235

neutronc:
Now also Tiledictionary


erik-vos (changes requested):
TileDesigner.18t also needs to be updated.

erik-vos (approved):
OK now!

PR #321, opened 2020-12-21
Fix for #320 and other issues

### LINKED ISSUE #320: 1835 - PR loses director
In certain situations where the PR director sells one or more share the director "disappears" from the game.
By disappears I mean that no player receives the director share and the game breaks.
The effect has been discovered in the current version (2.3.0) of Rails.

## Error Reproduction
1)  load: [1835_20201220_2109_Marc.rails.zip](https://github.com/Rails-18xx/Rails/files/5724383/1835_20201220_2109_Marc.rails.zip)
2)  sell 40% PR shares as player "Marc"
3)  now none of the players has the PR director

erik-vos:
Part of the work to (re)visit emergency train buying and bankruptcy follow-up.

- 18Scan bankruptcy now works
- Various 1835 issues fixed around share selling. Perhaps provisionally, work continues. Share selling seems to need an overhaul.
- Four test cases added for the above-mentioned issues.
- Issue #320 fixed.

Closes #320


src/main/java/net/sf/rails/game/GameManager.java
```diff
@@ -1295,8 +1299,63 @@ public void registerCompanyBankruptcy (PublicCompany company) {
         or.finishTurn();
     }
 
+    /**
+     * Process the effects of a player going bankrupt, without ending the game.
+     * This code applies to (most of) the David Hecht games.
+     * So far 1826, 18EU, 18VA, 18Scan have been identified to use this code.
+     */
     protected void processPlayerBankruptcy() {
```
madoar, line 1307: I personally don't like the idea to provide a default implementations that is used by a subset of all implemented variants.
This is not really targeted only to this method but to most other classes as well, where we have similar issues, because we implement a default behaviour based on one game variant (normally 1830) and then "change" it in a subvariant.
This makes it quite easy in my opinion to introduce bugs because the "fallback" behaviour, which is called when you don't provide a dedicated behaviour, is always the one from 1830.

Ideally I think we should make the behaviour in the variants more explicit. Maybe we could use a [mixin approach](https://en.wikipedia.org/wiki/Mixin)? In this case be case we could try to define an "empty" interface containing the method stub(s), an "default" interface containing the default implementation(s) and a number of differing implementations in the non-default variants. 

Alternatively we could think about externalizing the default behaviour in utility classes, which we can then explicitly call in all variants.

src/main/java/net/sf/rails/tools/ListAndFixSavedFiles.java
```diff
@@ -512,6 +506,37 @@ PossibleAction processInput() {
         }
     }
 
+    private class BuyCertificateDialog extends EditDialog {
```
madoar, line 509: I think this class should be moved to its own `.java` file. The file is becoming a bit too long

src/main/java/net/sf/rails/ui/swing/ORUIManager.java
```diff
@@ -1428,9 +1428,13 @@ public void updateStatus(PossibleAction actionToComplete, boolean myTurn) {
 
             // TEMPORARY extra message about having no route
             for (BuyTrain bTrain : possibleActions.getType(BuyTrain.class)) {
-                if (bTrain.isForcedBuyIfNoRoute()) {
+                if (bTrain.isForcedBuyIfHasRoute()) {
```
madoar, line 1431: The performed check does not fit with the comment above. Is this change intentional (I do not really know the performed logic here just want to be sure so that we don't introduce an error)?

src/main/java/net/sf/rails/ui/swing/ORUIManager.java
```diff
@@ -1428,9 +1428,13 @@ public void updateStatus(PossibleAction actionToComplete, boolean myTurn) {
...
                     b.append("<br><font color=\"red\">");
-                    b.append(LocalText.getText("MustBuyTrainIfNoRoute"));
+                    if (bTrain.isForcedBuyIfNoRoute()) {
```
madoar, line 1433: Isn't this check semantically contrary to the one above (i.e. `bTrain.isForcedBuyIfHasRoute()`)?

src/main/java/rails/game/action/BuyCertificate.java
```diff
@@ -41,6 +41,10 @@
     protected int price;
     protected int maximumNumber;
 
+    // In exceptional cases, a president's share may end up in the pool and be buyable.
+    // This can occur after a bankruptcy in 1835 and 18Scan
+    protected boolean president = false;
```
madoar, line 46: If the field can be accessed by a setter and getter method I think it should be made private

src/main/java/net/sf/rails/game/model/PortfolioModel.java
```diff
@@ -245,9 +247,9 @@ public int getShare(PublicCompany company) {
 
 
     /**
-     * @return the number of shares owned by the PorfolioModel for this company
+     * @return the number of shares owned by the PortfolioModel for this company
      */
-    public int getShareNumber(PublicCompany company) {
+    public int getShares(PublicCompany company) {
```
madoar, line 252: This name seems misleading to me. This method does not return the shared but the number of owned shares. I.e. I think it should be named something like `getNumberOfShares` right?

erik-vos: There are no share objects (but certificates), so you can't get any. Of course, what you propose is more unambiguous, but I consider brevity as a worthy goal too.

I tend to use the following nomenclature:
Share = a percentage of company ownership,
Shares = number of share units (usually 5% or 10%, sometimes more),
Cert = a single share certificate,
Certs = a collection of share certificates.

src/main/java/rails/game/action/BuyCertificate.java
erik-vos: I would prefer to give subclasses direct access.
Unfortunately, Java has no keyword for such access that excludes other classes in the same package.

src/main/java/net/sf/rails/tools/ListAndFixSavedFiles.java
erik-vos: Generally, I don't bother much with the length of java files. 
In this case, we are talking about a simple tool that does not even has its own package, but coexists with more such tools. Splitting off separate .java files would be confusing, unless we create separate packages for such stuff.
For me, this has priority zero.

neutronc (approved):

src/main/java/net/sf/rails/ui/swing/ORUIManager.java
erik-vos: The comment is old, I have not looked at it, other than recognizing that TEMPORARY means that automatic having-a-route checking has not yet been implemented.

Let's call the two conditions:
A = must have a train if the company has a route.
B = must have a train if the company has no route.
Now if B is true, A is always true as well, but if A is true, B can have both values.
So if A is true, a message is displayed, the contents of which depends on B. 
This is the intended, and I believe also actual, effect of this nested if construct.

src/main/java/net/sf/rails/game/GameManager.java
erik-vos: I could not immediately find examples of mixin constructs in Java, so I don't really get what you mean. 

Would your "default" interface implement the "empty" interface and the main (not game-specific) class, such as OperatingRound,  implement the default interface? Wouldn't then the default methods also extend to all games that don't override these? In that case, I don't see the difference with having stubs, or default methods, in the main class.

Or do you intend to restrict such interfaces to games that show a certain behaviour that does not apply to other games?
We could think of bankrupty processing, which does not apply to all games where a bankruptcy ends a game. The other games could then be the only ones implementing such interfaces. This could be a sensible idea.

In Rails 1.x we had, on someone's proposal, additional interfaces for all kinds of objects. Train and TrainI, Tile and TileI and so on and so on. These were hardly used, and thankfully removed in Rails 2.x.

Keeping the explosion of separate Java files under control is also a worthy goal to me.

src/main/java/net/sf/rails/ui/swing/ORUIManager.java
erik-vos: No, see my other comment on this class.

madoar (approved):

PR #322, opened 2020-12-22
enhancement: update to JDK 15 in pipeline

madoar:
This PR is blocked by a compatibility issue with the jacoco plugin


neutronc (dismissed):

neutronc:
Schau mal bitte wo der Ubuntu Fehler herkommt @madoar

madoar:
> Schau mal bitte wo der Ubuntu Fehler herkommt @madoar

The reason for the failing build is because of an incompatibility of the jacoco plugin we are using. Jacoco does not seem to be compatible with Java 15 as of now. For more details see the discussion in https://github.com/gradle/gradle/issues/15038

erik-vos:
And, by the way, is there anything useful in Java 15 that we are eager to use?

FYI, my Java knowledge does not go much farther than most of Java 5 and some bits of Java 8 (and does not include foreign packages from google or wherever).

jklap:
Maybe Records in 15? But certainly not worth moving to 15 just to try to force that use of them ;)

It's probably worth trying to stay semi-current with the JDK such to stay on top of bug fixes, byte-code optimizations BUT keeping the target compile as 11 or 14, whatever we had decided previously. While I personally keep 8, 11, 14 and now 15 installed locally I find that I only ever use the most recent JDK...

madoar:
The main reason for the update here is that Java 14 is not supported anymore. It only had short-term support for 6 months which stopped with the release of Java 15.

madoar:
@neutronc I think we can merge this now.

What worries me a little is that the `Create Deb Installer` and `Create Rpm Installer` jobs contain some `java.io.IOException` exceptions because of missing system libraries. For some reason this does not lead to a failing execution of `jpackage`, instead both a `deb` and `rpm` installer are created. This also happened for the latest tag (i.e. version `2.4.5`), therefore I think the exceptions are unrelated to this PR.

madoar:
In the future we will need a followup PR with an update to Java 16, which has been released a week ago.

PR #323, opened 2020-12-22
enhancement: update gradle configuration

madoar:
This PR fixes the trailing `+` bug in the splash screen.
In addition it updates some parts of the gradle configuration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -143,15 +143,23 @@ dependencyUpdates {
     }
 }
 
+/**
+ * Converts the version to a jpackage conform form.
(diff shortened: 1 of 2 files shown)


neutronc (approved):

jklap (approved):

madoar:
@neutronc because of the windows version limitation in the form of `major.minor.bugfix` we can't encode information about alpha, beta and final release versions in the package version. Therefore I remove this information from the version string. I hope this approach is fine with you?

PR #324, opened 2021-01-03
Pr mbr 18 chesapeake

neutronc:
18 Chesapeake : Current State no multiple Tile lays for Privates implemented yet. Rest should be playable,


erik-vos:
I don't know this game, so I can't say much about it.

Just one remark, after a blink at the rules: the multiple tile laying privates seem much alike 1835 OBB, which has two separate properties for the two locations, apart from the (configurable) ownership and the connectedness requirement; only the latter one presents a real problem IMHO. Can't you use that scenario?

erik-vos (approved):

neutronc:
> 
> 
> I don't know this game, so I can't say much about it.
> 
> Just one remark, after a blink at the rules: the multiple tile laying privates seem much alike 1835 OBB, which has two separate properties for the two locations, apart from the (configurable) ownership and the connectedness requirement; only the latter one presents a real problem IMHO. Can't you use that scenario?

Can the OBB lay both tiles in the same turn ?

erik-vos:
No, but that is excluded in the code: OperatingRound_1835.

erik-vos:
The connectedness check between the two tiles IMO might be doable using the kind of bit manipulation I have been doing in MapManager.findNewBridgeSides() for SOH, using HexSideSet bitmaps.

PR #347, opened 2021-02-20
Finalizing 18Chesapeake for Beta Test

neutronc:
!8Chesapeake should be fully playable now, could use some test parties.


erik-vos:
I cannot really judge this, as I don't know anything about 18Chessie, but I haven't seen anything that did raise my eyebrows, so let's get on.

Issue #376, opened 2021-03-27
1837: Mine revenues of EPP and SPB are incorrect


erik-vos:
@hast18xx reported in #346:

[18] Kohlebahn values should be corrected (because my version1 map says so, I hope this is still valid)
E11 EPP is 30,50  should be  20,40
K31 SPB is 20,50 should be 30,50

erik-vos:
Actually, the (almost) correct revenues were there, but the v1 and v2 versions were the wrong way around.
Also, the Romoth variant amounts were equal to the v2 values, but in all  likelyhood these should be the v1 values.
Fixed.

Issue #375, opened 2021-03-27
1837: Sold-out company where president has <= 40% shares does not rise at end of SR


erik-vos:
In #346 @jdgalt reported:

[17a] > it is SR end and all shares are in players hand
> in case of 50%+ it is ok (up and one to the left)
> in case of 40% the price remains but should move up and one to the right
> 
> ![SR end shift](https://user-images.githubusercontent.com/64490299/110147172-a4230380-7ddb-11eb-8d08-2b3c0dc195dd.png)

This seems to me to be a contradiction.

Remember that the actual published game uses a hex grid for the stock market (photo attached).  The translation of that into a square grid seems to be that the upper rows are shifted a half-space to the left per row.

Thus you're correct that the movement for "all sold out, president has 50% or 60%" is up-and-left (diagonal) in both the hex grid and the square grid versions.  But the movement for "all sold out, president has 40% or less", which was up-and-right in the hex grid version, is straight up in the square grid version.

![1837_market_snapshot](https://user-images.githubusercontent.com/9399882/110176287-e9353e80-7db7-11eb-916f-5211838427fd.jpg)

and @hast18xx chimed in:

[17b] > > it is SR end and all shares are in players hand
> > in case of 50%+ it is ok (up and one to the left)
> > in case of 40% the price remains but should move up and one to the right

> 
> This seems to me to be a contradiction.
> 
> Remember that the actual published game uses a hex grid for the stock market (photo attached). The translation of that into a square grid seems to be that the upper rows are shifted a half-space to the left per row.
> 
> Thus you're correct that the movement for "all sold out, president has 50% or 60%" is up-and-left (diagonal) in both the hex grid and the square grid versions. But the movement for "all sold out, president has 40% or less", which was up-and-right in the hex grid version, is straight up in the square grid version.

yes you are right
so it should be like this
 in case of 50%+ it is ok (up and one to the left)   
 in case of 40% the price remains in same column but should move UP   

BK open at 70 , no engine  left to 67 and after the next SR up and left to 69
![lemmy1](https://user-images.githubusercontent.com/64490299/110201632-63bb9800-7e64-11eb-8d7a-57a88d0d9619.PNG)

BK open at 70 , no engine  left to 67 and after the next SR up  to 72
![lemmy2](https://user-images.githubusercontent.com/64490299/110201634-68804c00-7e64-11eb-9942-e11a7ab6e0c7.PNG)
 

need correction for BH
[...]

Issue #374, opened 2021-03-27
1837: More than 2 companies can have same par price


erik-vos:
Reported by @hast18xx in issue #346:

[14] f) each par value can only be used twice  .  Here:  85 has been used three times.  See SC

![par issue](https://user-images.githubusercontent.com/64490299/109968533-21267e00-7cf3-11eb-9c21-241f211de0bf.png)

erik-vos:
Fixed.

Issue #373, opened 2021-03-26
1837: Price drops one row for each sold share.


erik-vos:
@hast18xx reported in issue #346:

[15] g) while selling shares price dropping is wrong. Now it is like 1830 but it should be only one drop per action, like in 1835

Will look at this one now.

PR #377, opened 2021-03-27
1837: Fixed financial pre-4-train omissions and bugs

### LINKED ISSUE #359: 1835: Layout ISR window is weird
Rails version 2.4.5, 1835, Clemens variant.

The layout of the ISR window is weird. In earlier versions, the available certificates were just listed top to bottom. In this version there is first a list of four certs, then next to it a list of three certs, etc. 
Is this something intended, or did I do something wrong?

erik-vos:
Bug fixes:

- Enabled 'Split' option for major company revenue
- Fixed rounding issues
- Fixed price dropping on selling more than one share
- Implemented max. 2 companies per par price
- Fixed price changes of sold-out majors
- Fixed incorrect standard mine revenues

Improvements:

- Generic rounding methods, also applied to 18EU
- Implemented the remaining aspects of the Romoth variant

Closes #359, #361, #372, #373, #374, #375, #376


erik-vos:
All installer tests of this PR fail, so something must be wrong. I cannot research that, but I got a glimpse of what looked like a problem with Java 15. Is that update incomplete, or bad, or what? Can we revert?

neutronc:
The installer tests were successfull for that commit. So something changed with your PRs need to investigate what happened there.

erik-vos:
Some regression tests in the master branch currently fail. Do the installation tests repeat these tests? If so, the  cause is clear.

erik-vos:
OK, I have removed the 4 failing saved test files and their reports.

Two of these are 1837 saved files that were known to be failing and will have to be created anyway
The other two were:
- 18Scan_AfterBankruptcy1.rails 
- 1880_ATG1
These failures is very strange, because AFAIK nothing significant has been changed in the generic code.

In the 18Scan case, the problem seems to be, that a company gets a turn although it should not. It is 'hibernating' (i.e. inactive) because its president has gone bankrupt.

I have not looked at the 1880 case; I will check it, but I know very little about this game, so maybe I will have to delegate this one.

Anyway, all installer tests now succeed, so I suppose this PR is ready to be merged.

src/main/java/net/sf/rails/game/OperatingRound.java
```diff
@@ -2757,6 +2767,106 @@ public void splitRevenue(int amount) {
 
     }
 
+    /*
+     * =======================================
+     *  6.4. ROUNDING
```
neutronc, line 2772: Good Idea

neutronc (approved):

erik-vos:
After a long day of debugging I found the small error that inhibited an 18Scan company to be restarted after its presidency got into the Pool after a bankruptcy. Miraculously, this also fixed the 1880 saved file problem.
Also StockRound has been updated to add an overlooked option to restart such a company.
I'll make a new PR for this one.

madoar:
All GitHub Actions pipelines call `./gradlew clean test`. Most pipelines also call `./gradlew clean build`, which also contains a `./gradlew test` call. I'm not entirely sure how the tests for Rails are configured but I imagine they are called entirely

PR #434, opened 2021-11-08
Permitting tile lays on reserved hexes by other players

erik-vos:
If the current player also owns the company for which the hex is reserved, the permission is automatic.

If not, the current player will be asked if the other player has given permission (off-game). Honesty is assumed here.

As a last resort, reserved hexes can now also be laid in correction mode.


src/main/java/net/sf/rails/game/OperatingRound.java
```diff
@@ -1622,16 +1622,6 @@ public boolean layTile(LayTile action) {
...
-            /*
-            if (stl != null && stl.isFree()) {
-                cost = 0;
-            } else {
-                cost = hex.getTileCost() + extraLayTileCost(action);
-                if (stl != null) {
-                    cost = Math.max(0, cost - stl.getDiscount());
-                }
-            }
-            */
```
neutronc, line 1634: Where is that check done now ? In TileLayCost() i assume ?

erik-vos: No. The initial check is done in a new method gotPermission(), which is called from hexClicked(). The answer is handled in dialogActionPerformed(). All in the same class ORUIManager.

neutronc: And there you are and  i missed that comment block, sorry.

neutronc (approved):

PR #452, opened 2022-01-23
18EU: three variants added

erik-vos:
1. Two options for a 16th minor with 4 players: 0 on Hamburg or 16 on Frankfurt
(There is no check on the actual number of players)

A new underlying Hamburg tile -1903 with a one-slot city is used if the Hamburg option is chosen. A new value ALWAYS of Access.RunThrough is used to allow a Hamburg pass-through even if it is tokened.

2. An option to allow minors upgrade their home base to green from phase 3.
And added the new tile -1903, which I had overlooked in the first commit.

3. David Hecht's Alpine variant.


erik-vos:
Not sure why these checks fail. It can hardly be my latest commit, which was tiny.
The error messages point to Gradle, which is out of my scope.

If I only knew how to rerun these checks...

erik-vos:
Since about two weeks all new PRs fail on the MacOS installer.

Can please some knowledgeable person look into this problem?

The failed check reports include a lot of info, but what strikes me is the following message, which appears just before the Exceptions start:
  `hdiutil: unflatten: verb not recognized`
Don't know if this is the bug.

erik-vos:
Also added a fix to the remaining 1835 OBB closure issues.

erik-vos:
After a bit of googling I found that `hdiutil flatten/unflatten` (whatever that means) is no longer supported by MacOS Big Sur.
My conclusion is, that the problem is internal to the MacOS installer check procedure, and has nothing to do with Rails code.

My question: is it possible to disable or bypass that particular check, as long as we haven't found anyone to fix it? For now, it's just blocking progress without any sight on fixing it.

neutronc:
It should be fixed by my latest commit.

PR #459, opened 2022-02-13
Two minor improvements

erik-vos:
1. Four almost identical discardTrain() methods have been merged into one: an instance method of the DiscardTrain class.
It may be a novelty to have an action object process itself, but in view of the coding virtue of encapsulation it seems like a logical thing to do.

2. Invisible underlaying tiles did show up in the Remaining Tiles panel. This has been fixed.


erik-vos:
All installer checks have passed. This time, I have merged this PR myself.

PR #471, opened 2022-06-21
1826: added implementation details (1), and several fixes.

erik-vos:
1826 is now playable through phases 2H and 4H. The main missing feature is that token lay cost is still calculated "as the crow flies".

The main additions are:
- All special properties work.
- Mechanism to create extra base tokens (list of all tokens is no longer "immutable").
- Mechanism to "grow" companies, i.e. add extra shares (5-share to 10-share). Numbers still hardcoded.
- Reaching destination enables growing. Both are manual actions on the Special menu.
- Changeable (and stateful) train limits, as these change when growing.

Minor details:
- Fixed and much simplified ArrayListState.setTo(). It crashed on IndexOutOfBounds if the new list was 2 items shorter than the old one, as appeared to occur with the 1837 Ux minors.
- Fixed the appearance and contents of the Rights column in ORPanel (also used in 1830 Coalfields).


erik-vos:
Also fixed share selling in 1835, discussed at length in issue #475. From now on, only _whole_ certificates can be sold, and only one size (10% or 20%) at a time. By default, multiple sales of the same company is done at the same price. This behaviour can be changed via unchecking an Option at game start, or  via the Special menu _immediately_ after a second (or third, etc.) consecutive sale. 
`StockRound.setSellableShares()` has been renamed to `setSellShareActions()` and was partly rewritten, also the overriding method in `ShareSellingRound` has been merged into it.

I have added the 1835 changes into this PR because the latter appeared to be still open. If nothing happens within a week, I will merge the whole thing myself.

erik-vos:
I have added the "greediness" fix described in issue #478. See there for further information.

erik-vos:
Closes #478

PR #494, opened 2022-10-16
Creating new Version v 2.5

neutronc:


erik-vos (approved):
Actually, the main achievement of this release is the completion of 1837 (still marked beta though).
1826 is being worked on (I'm currently wrestling with the hex distance to determine tokening cost to be counted via laid track. A few weeks ago it seemed to work, but later appeared to break with closed loops.)

Issue #520, opened 2023-02-23
1826: Non-optimal runs for 10H-trains


erik-vos:
During testing using a real game (published by Steve Thomas in his 'zine) I noticed that the revenue calculator sometimes does not find the best run for a company. Examples:

![1826-GCmissingM13](https://user-images.githubusercontent.com/13908049/220969195-b1837af6-4fa0-46fe-a7c7-cfb3a210db88.jpg)
The GC run with a 10H-train should have included M13 (St. Etienne) for an F30 higher income.

![1826_AlsMissingK17](https://user-images.githubusercontent.com/13908049/220969679-8120f21a-36f6-447a-8174-30d244838691.jpg)
The Alsatian run with a 10H-train should have included K17 (Geneva) for an F30 higher income.

Possibly this bug is specific to H-trains, as in my experience most 'normal' train runs are OK.
The bad news is that I have little hope to be able fix this problem - the revenue calculator still is much of a mystery to me.

I have added an advice to the 1826 game notes to enable the revenue spinner, and correct non-optimal runs.

erik-vos:
It turns out that the lengths of H-trains were not always taken into account in the route finding process for these trains. The whole H-train route calculation process uses the variable `maxMajors`, which gets a different meaning with H-trains. It seems to me that overriding `RevenueCalculatorMulti` methods in `RevenueCalculatorMultiHex` is incomplete, Maybe the former class should get an additional separate `maxHexes` attribute per train, which is used everywhere where currently `maxMajors` and `maxMinors` are used with a H-train. But currently I have no grip on all that stuff.

To the defence of the developer of this code (AFAIK  that was Stefan) I would say, that 1826 probably is the first game in Rails that uses H-trains, so I may be the first one who is really seriously testing with that type of train.

I have added a few stopgaps in `RevenueAdapter` and `RevenueCalculator` to make it work _most of the time_, but there still are calculations that go wrong, sometimes even resulting in a zero revenue. So my recommendation to use the revenue spinner still stands.

PR #521, opened 2023-02-23
Bug fixes for 1826, implemented special revenue details for 18VA

erik-vos:
During replaying a real 1826 game a lot of bugs and missing details have been detected and fixed. A problem remains with the calculation of H-train revenues, in particular 10H-trains, where sometimes cities are bypassed; see issue #520 and the 1826 game notes.

Further 18VA details have been added, in particular the various off-map bonuses. Special code had to be written to correctly handle 'minor' cities. The 'minor' attribute, which ever was only intended to facilitate n+m-trains, is now also used to determine which stops should be ignored with certain trains, and that is a matter of all or nothing. But in 18VA passenger trains ignore mines (represented by towns), but honour ports, although neither count against train length. I have been trying to refine the specification of such things, but that will affect the revenue calculation code with which I am not familiar.


erik-vos:
The Ubuntu installer check fails on this PR. Looking back into Slack, I notice that this has happened several times since 13 dec 2022 with PRs from dependabot. In some cases the problem seems to have been fixed by (automatically?) deleting something. Can someone please sort out what is going on, and fix it?

erik-vos:
Merged even though the Ubuntu Installer check still fails

PR #534, opened 2023-04-28
18VA near completed

erik-vos:
Added default revenue calculation methods to RevenueManager, and a subclass RevenueManager_18VA to provide real stop values for the initialization of each NetworkVertex, possibly depending on train type and company properties. So RevenueManager is no longer a final class. IMO this makes the predictionValue() method in dynamic modifiers obsolete.

Also added a simple Revenue class, that carries two integers for normal and special revenue. For 18VA, special is used for the mine income that goes directly to the company treasury (as 1837). The intention is to have revenue objects carry all revenues, but for now it is only applied to 18VA.


erik-vos:
This time **all** checks fail. No clue why. Could one of the 'dependabot' PRs be the cause? 
Unfortunately, that stuff is out of my scope.

